PR #122, opened 2014-12-16
Make the db URL configurable

sbrunner:
requires: https://github.com/camptocamp/c2cgeoportal/pull/1161#issuecomment-67154926

I put an arbitrary my_maps URL !

### DIFF
--- a/vars_geoportailv3.yaml
+++ b/vars_geoportailv3.yaml
@@ -22,6 +22,9 @@ vars:
     # closure paths
     closure_library_path: 'process.stdout.write(require("openlayers/node_modules/closure-util").getLibraryPath())'
 
+    servers:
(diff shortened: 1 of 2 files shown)


vars_geoportailv3.yaml
```diff
@@ -22,6 +22,9 @@ vars:
     # closure paths
     closure_library_path: 'process.stdout.write(require("openlayers/node_modules/closure-util").getLibraryPath())'
 
+    servers:
+        my_maps: http://map.geoportail.lu/my_maps
```
jaykayone: this should be http://ws.geoportail.lu/mymaps (cf https://github.com/Geoportail-Luxembourg/geoportailv3/issues/131)

PR #126, opened 2014-12-17
mako template with proxy and logging settings

jaykayone:
this is a pull request with a questionmark.

to make the solution work in our wsgi environment, we have to define the proxy in the wsgi config.
I am not sure though if this has to be committed into master.
should this proxy setting be included in the makefile / vars-file rather than in the mako template directly?

### DIFF
--- a/apache/geoportailv3.wsgi.mako
+++ b/apache/geoportailv3.wsgi.mako
@@ -1,11 +1,13 @@
 from pyramid.paster import get_app
+
+%if http_proxy:
 import os
-os.environ['http_proxy'] = 'http://proxy:3128'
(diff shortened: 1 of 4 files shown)


elemoine:
> should this proxy setting be included in the makefile / vars-file rather than in the mako template directly?

Yes, I think so.

You could add `http_proxy` to your yaml config file and do something like this in the mako file:

```
% if http_proxy:
import os
os.environ['http_proxy'] = '${http_proxy}'
os.environ['https_proxy'] = '${http_proxy}'
% endif

Does it make sense?

elemoine:
Please merge when you think this is ready.

elemoine:
I assume it works as expected when there's no `http_proxy` defined.

PR #134, opened 2015-01-06
More specific pyramid_closure root for ngeo

elemoine:
This PR uses a more specific pyramid_closure `root` for ngeo. This is to avoid problems when using a built ngeo source tree (with `ngeo.js` and `ngeo-debug.js` in `dist/`).


pgiraud:
Please merge.

PR #136, opened 2015-01-06
Layer factory refactoring

elemoine:
This PR is a refactoring of the layer factory code. There are now more services. And the `appGetWmtsLayer` service is now used in the main controller to create the background layer. Later, the `appGetWmtsLayer` will be used by the background layer component.

Please review.

### DIFF
--- a/geoportailv3/static/js/layerfactoryservice.js
+++ b/geoportailv3/static/js/layerfactoryservice.js
@@ -1,3 +1,5 @@
+goog.provide('app.GetWmsLayer');
+goog.provide('app.GetWmtsLayer');
 goog.provide('app.LayerFactory');
 
 goog.require('app');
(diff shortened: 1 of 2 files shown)


geoportailv3/static/js/layerfactoryservice.js
```diff
@@ -9,116 +12,164 @@ goog.require('ol.tilegrid.WMTS');
...
+    var type = node['type'];
+    var layerCacheKey = type + '_' + node['name'];
+    if (layerCacheKey in app.layerCache_) {
+      return app.layerCache_[layerCacheKey];
+    }
+
+    var layer;
+
+    if (type.indexOf('WMS') != -1) {
+      layer = appGetWmsLayer(node['name'], node['name'], node['url']);
```
pgiraud: I don't see any reason why `name` and `label` would be different.

pgiraud:
I added one comment. Otherwise, looks good to me.

geoportailv3/static/js/layerfactoryservice.js
elemoine: You're right. I removed the `label` argument.

elemoine:
For the record, I will need to refactor the layer factory services again for the background layer component, but this is a start, and this also fixes a few bugs I've seen while doing the refactoring.

PR #139, opened 2015-01-07
Languages2

petzlux:


petzlux:
Jquery doesnt get downloaded during build ? Anyone know why ?

elemoine:
> Jquery doesnt get downloaded during build ? Anyone know why ? @elemoine

With the master branch jQuery does get downloaded for me when I run the `build` target.

petzlux:
@sbrunner Jeff me disait que t'avais des commentaires concernant les changements de CONST_Makefile ?

sbrunner:
Hello,

Yes, some task like `update-translation` is luxembourg specific than it shouldn't be in the `CONST_Makefile`, but actually we can't move it properly to the `geoportailv3.mk`, then I will move it later :-)

petzlux:
Ready for merge ?

geoportailv3/utils/db_to_pot.py
```diff
@@ -0,0 +1,22 @@
...
+from os import path
+import codecs
+
+
+pathToPotFiles = './geoportailv3/locale/'
+p = path.join(pathToPotFiles, "geoportailv3-db.pot")
+
+if __name__ == "__main__":
+    w = codecs.open(p, 'w', encoding='utf-8')
+    w.write('#, fuzzy\nmsgid ""\nmsgstr ""\n"Plural-Forms: nplurals=2; \
```
sbrunner: for me the 
```
#, fuzzy
msgid ""
msgstr ""
```
looks strange Is it relay needed ?

sbrunner:
Just a last little question, otherwise that's looks good :-)

CU
Stéph

geoportailv3/utils/db_to_pot.py
petzlux: AFAIK, it is part of the specification of a pot / po file, and gettext tools will fail if not present.

sbrunner: OK, thanks for the verification :-)

petzlux:
If I want to merge into master, than will Pierrre's Commits here appear twice in master ? It looks this way in my git log at the moment , so I dont want to do a push

elemoine:
Sounds like some rebase is needed. I'll look at it later today or tomorrow morning if you don't feel comfortable with the required operation.

petzlux:
after a couple of unsuccessful tries, #149 seems to look better and can be merged without conflicts !

PR #142, opened 2015-01-07
Layer manager

pgiraud:


geoportailv3/static/js/layermanager/layermanager.html
```diff
@@ -0,0 +1,15 @@
+<div ng-if="!layermanagerCtrl.layers.length" translate>No layer selected</div>
+<ul id="layermanager-{{::layermanagerCtrl.uid}}" class="layermanager">
+  <li ng-repeat="layer in layermanagerCtrl.layers" class="panel">
+    {{ layer.get('label') }}
+    <a data-toggle="collapse" data-parent="#layermanager-{{::layermanagerCtrl.uid}}" href="#layermanager-item-{{ $index }}">
```
elemoine: You sometimes use no spaces within `{{ }}` and sometimes you do.

elemoine: And FWIW I don't think we need spaces.

geoportailv3/static/js/layermanager/layermanager.html
```diff
@@ -0,0 +1,15 @@
...
+      <span class="collapsed fa fa-ellipsis-h"></span>
+      <span class="expanded fa fa-remove"></span>
+    </a>
+    <div id="layermanager-item-{{ $index }}" class="collapse">
```
elemoine: Shouldn't we use the uid trick here again?

geoportailv3/static/js/layermanager/layermanagerdirective.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+app.layermanagerDirective = function(appLayermanagerTemplateUrl) {
+  return {
+    restrict: 'E',
+    scope: {
+      'map': '=appLayermanagerMap',
+      'layers': '=appLayermanagerLayers'
+    },
+    controller: 'AppLayermanagerController',
+    controllerAs: 'layermanagerCtrl',
```
elemoine: I don't think we need "complex" name for directive controllers. The name is local to the directive, so I think that we can just use "ctrl".

geoportailv3/templates/index.html
```diff
@@ -51,18 +51,25 @@ <h2 translate>Layers</h2>
             ✕
           </button>
           <ul class="nav nav-tabs nav-justified">
-            <li role="presentation">
-              <a href="#mylayers" ng-click="mainCtrl.showTab($event);">My Layers</a>
-            </li>
             <li role="presentation" class="active">
-              <a href="#catalog" ng-click="mainCtrl.showTab($event);">Catalog</a>
+              <a href="#mylayers" data-toggle="tab">My Layers
+                <span ng-if="mainCtrl.selectedLayers.length > 0">({{ mainCtrl.selectedLayers.length }})</span>
```
elemoine: You use spaces after `{{` and before `}}` here again.

geoportailv3/static/js/maincontroller.js
```diff
@@ -86,6 +90,16 @@ app.MainController.prototype.setMap_ = function() {
 
 
 /**
+ * @private
+ */
+app.MainController.prototype.updateSelectedLayers_ = function() {
+  var items_ = this['map'].getLayers().getArray().slice().reverse();
+  items_.pop();
+  this['selectedLayers'] = items_;
```
elemoine: This code deserves a comment (in the code) I think. The call to `pop` is obscure if you don't know that we want to exclude the background layers here.

elemoine: I also note that we will need a more complex filter when we'll use vector layers for drawing etc.

geoportailv3/static/js/layermanager/layermanager.html
pgiraud: No. I don't think it's worth since it's a flat array.

pgiraud:
I took Éric's comments into account. I'm now waiting for travis tests to pass and will merge if everything goes well.

PR #151, opened 2015-01-16
Prevent injected service from being renamed

pgiraud:
Fixing the issue with layer manager not showing on built version.
Please review.


elemoine:
Yes, that's it!

PR #154, opened 2015-01-16
Use layer expr in catalog tree partial

elemoine:
This is to accommodate for the changes in https://github.com/camptocamp/ngeo/pull/140.


elemoine:
Note that this should be reviewed and merged before doing more work on the catalog.

elemoine:
PR updated based on the latest changes in ngeo.

elemoine:
Merging this to remove the `TypeError` that we currently get when selecting a layer in the catalog.

PR #155, opened 2015-01-16
Opacity and visibility interaction

pgiraud:
With this pull request, the opacity and visibility now work together.
Moving the opacity slider to the left will display a slashed eye.
Clicking on the eye will change the opacity to 0 and thus will move the slider automatically to the left.
When the slashed eye is clicked:
 - the opacity is changed to 1 if the slider was manually moved to the left,
 - the opacity is changed to the previously known opacity if the visibility was changed by clicking on the eye.

### DIFF
--- a/geoportailv3/static/js/layermanager/layermanagerdirective.js
+++ b/geoportailv3/static/js/layermanager/layermanagerdirective.js
@@ -60,4 +60,13 @@ app.LayermanagerController.prototype.removeLayer = function(layer) {
 };
 
 
+/**
+ * @param {ol.layer.Layer} layer Layer.
+ * @export
+ */
+app.LayermanagerController.prototype.changeVisibility = function(layer) {
+  layer.set('opacity', layer.get('opacity') > 0 ? 0 : 1);
+};
+
+
 app.module.controller('AppLayermanagerController', app.LayermanagerController);
--- a/geoportailv3/static/js/layermanager/layermanagerdirective.js
+++ b/geoportailv3/static/js/layermanager/layermanagerdirective.js
@@ -65,7 +65,17 @@ app.LayermanagerController.prototype.removeLayer = function(layer) {
  * @export
  */
 app.LayermanagerController.prototype.changeVisibility = function(layer) {
-  layer.set('opacity', layer.get('opacity') > 0 ? 0 : 1);
+  var currentOpacity = parseFloat(layer.get('opacity'));
+  var newOpacity;
+  if (currentOpacity === 0) {
+    newOpacity = layer.get('oldOpacity') ? layer.get('oldOpacity') : 1;
+    // reset oldOpacity for later use
+    layer.set('oldOpacity', undefined);
+  } else {
+    layer.set('oldOpacity', currentOpacity);
+    newOpacity = 0;
+  }
+  layer.set('opacity', newOpacity);
 };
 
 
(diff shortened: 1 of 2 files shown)


geoportailv3/static/js/layermanager/layermanagerdirective.js
```diff
@@ -45,4 +45,23 @@ app.LayermanagerController.prototype.removeLayer = function(layer) {
 };
 
 
+/**
+ * @param {ol.layer.Layer} layer Layer.
+ * @export
+ */
+app.LayermanagerController.prototype.changeVisibility = function(layer) {
+  var currentOpacity = parseFloat(layer.get('opacity'));
```
elemoine: I'd use `layer.getOpacity` rather than `layer.get('opacity')` to make it clear that "opacity" is a built-in property as opposed to a user-defined property.

elemoine: Also, I don't think we should use a `parseFloat` here. This sounds like a bug in ngeo.

geoportailv3/static/js/layermanager/layermanagerdirective.js
```diff
@@ -45,4 +45,23 @@ app.LayermanagerController.prototype.removeLayer = function(layer) {
...
+  var newOpacity;
+  if (currentOpacity === 0) {
+    newOpacity = layer.get('oldOpacity') ? layer.get('oldOpacity') : 1;
```
elemoine: Are you testing for the case where the old opacity is `0` or `undefined` here? I think we should make the code for explicit. For example:

```js
    var oldOpacity = layer.get('oldOpacity');
    newOpacity = goog.isDef(oldOpacity) ? oldOpacity : 1;
    // …
```

geoportailv3/static/js/layermanager/layermanagerdirective.js
elemoine: Currently checking if there's a bug in ngeo…

elemoine: See https://github.com/camptocamp/ngeo/pull/146.

pgiraud:
Please review.

geoportailv3/static/js/layermanager/layermanagerdirective.js
```diff
@@ -45,4 +45,24 @@ app.LayermanagerController.prototype.removeLayer = function(layer) {
...
+  if (currentOpacity === 0) {
+    var oldOpacity = layer.get('oldOpacity');
+    newOpacity = goog.isDef(oldOpacity) ? oldOpacity : 1;
+    // reset oldOpacity for later use
+    layer.set('oldOpacity', undefined);
+  } else {
+    layer.set('oldOpacity', currentOpacity);
+    newOpacity = 0;
+  }
+  layer.set('opacity', newOpacity);
```
elemoine: Please use `setOpacity` here.

elemoine:
I've just added a comment. Otherwise looks good to me.

pgiraud:
Here's the closure compiler error:
```
ERR! compile /home/travis/build/Geoportail-Luxembourg/geoportailv3/geoportailv3/static/js/layermanager/layermanagerdirective.js:79: ERROR - actual parameter 1 of ol.layer.Base.prototype.setOpacity does not match formal parameter
ERR! compile found   : (Object|boolean|null|number|string)
ERR! compile required: (number|undefined)
ERR! compile   layer.setOpacity(newOpacity);
```

I'm pretty sure I have to add a @type definition somewhere.
But, for now, I didn't find any suitable fix for this error.

PR #156, opened 2015-01-16
Fix the scaleline directive

elemoine:
This is related to https://github.com/camptocamp/ngeo/pull/143.


elemoine:
Merging this to remove the `TypeError` that we currently get when loading the page.

PR #157, opened 2015-01-19
Add background layer selector

elemoine:
This PR adds a background layer selector based on Angular UI Bootstrap's `dropdown` directive and ngeo's `ngeoBackgroundLayerMgr` service.

This is what it looks like for the moment:

backgroundlayer

Work on the CSS is needed. In particular, the list is not correctly centered:

backgroundlayer-list

Things to do:

- [x] Decide if we use Angular UI Bootstrap (this implies changing other parts of the application)
- [x] Review and fix CSS (with @pgiraud)
- [ ] Rely on catalog web service instead of hard-coding the list of background layers

### DIFF
--- a/geoportailv3/static/js/backgroundlayer/backgroundlayer.html
+++ b/geoportailv3/static/js/backgroundlayer/backgroundlayer.html
@@ -1,0 +1,13 @@
+<div class="text-center">
+  <span translate>Background layer:</span>&nbsp;
+  <div class="dropdown">
+    <button type="button" class="btn btn-default" data-toggle="dropdown">
+      {{ctrl.bgLayer.name}} <span class="caret"></span>
+    </button>
+    <ul class="dropdown-menu" role="menu">
+      <li ng-repeat="layer in ::ctrl.bgLayers">
+        <a href ng-click="ctrl.setLayer(layer)">{{layer.name}}</a>
+      </li>
+    </ul>
+  </div>
+</div>
--- a/geoportailv3/static/js/backgroundlayer/backgroundlayerdirective.js
+++ b/geoportailv3/static/js/backgroundlayer/backgroundlayerdirective.js
@@ -1,0 +1,104 @@
+/**
+ * @fileoverview This file provides the "backgroundlayer" directive. This
+ * directive is used to create a dropdown for selecting the map's background
+ * layer. This directive is based on Bootstrap's "dropdown" component, and
+ * on the "ngeoBackgroundLayerMgr" service.
+ *
+ * Example:
+ *
+ * <app-backgroundlayer app-backgroundlayer-map="::mainCtrl.map">
+ * </app-backgroundlayer>
+ *
+ * Note the use of the one-time binding operator (::) in the map expression.
+ * One-time binding is used because we know the map is not going to change
+ * during the lifetime of the application.
+ */
+goog.provide('app.backgroundlayerDirective');
+
+goog.require('app');
+goog.require('app.GetWmtsLayer');
+goog.require('ngeo.BackgroundLayerMgr');
+
+
+/**
+ * @param {string} appBackgroundlayerTemplateUrl URL to backgroundlayer
+ *     template.
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+app.backgroundlayerDirective = function(appBackgroundlayerTemplateUrl) {
+  return {
+    restrict: 'E',
+    scope: {
+      'map': '=appBackgroundlayerMap'
+    },
+    controller: 'AppBackgroundlayerController',
+    controllerAs: 'ctrl',
+    bindToController: true,
+    templateUrl: appBackgroundlayerTemplateUrl
+  };
+};
+
+
+app.module.directive('appBackgroundlayer', app.backgroundlayerDirective);
+
+
+
+/**
+ * @constructor
+ * @param {ngeo.BackgroundLayerMgr} ngeoBackgroundLayerMgr Background layer
+ *     manager.
+ * @param {app.GetWmtsLayer} appGetWmtsLayer Get WMTS layer function.
+ * @export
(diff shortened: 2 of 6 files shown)


elemoine:
@sbrunner question for you: where do I find the list of background layers for a given theme in the response from the `themes` web service?

The `metadata` object includes a `bg_layer` property but I'm not sure this is what I'm looking for. My impression is that it is the theme's default background layer. No?

jaykayone:
there is a group called bg_layers
the one you found in the metadata is indeed the default one

elemoine:
http://localhost:5000/themes?min_levels=1&interface=desktop&catalogue=true&version=2 does not include the `bg_layers`.

jaykayone:
i think it should be like this : but it does not work now
i think c2cgeoportal is not uptodate in our solution

http://devv3.geoportail.lu/main/wsgi/themes?version=2&group=bg_layers&catalog=true

elemoine:
@sbrunner, should I use `group=bg_layers` in the the URL? Does that mean the background layer component needs to trigger a specific request?

sbrunner:
@elemoine yes :-)

elemoine:
> @elemoine yes :-)

This is a shame, no? It would be good to use one request only, don't you think?

sbrunner:
Yes it will be good,

We can ask with the query params:
* min_levels=1
* interface=desktop
* version=2
* baselayers_group=bg_layers
* catalog=true

And get the result:
```
{
    themes: [<themes>],
    baselayers: [<layers>],
    errors: [<errors>]
}
```
Does it looks good for you ?

elemoine:
> Does it looks good for you ?

Yep.

For consistency, I'd just use `bg_layers_group` as the request param (instead of `baselayers_group`), and `bg_layers` in the result (instead of `baselayers`).

elemoine:
@sbrunner, are you able to work on that?

sbrunner:
For consistency with what ? OL3 ?

sbrunner:
Yes but probably next week !

elemoine:
> For consistency with what ? OL3 ?

No, we already have `bg_layer` in the metadata.

elemoine:
```
{
    themes: [<themes>],
    baselayers: [<layers>],
    errors: [<errors>]
}
```

Today, we have `items` instead of `themes`. Do you also want to make that change?

Also, having the list of background layers at the first level in the JSON object means that we have no association between themes and background layers. Right?

sbrunner:
> No, we already have bg_layer in the metadata.

This metadata is completely free word for c2cgeoportal, and only for Luxembourg, should we relay follow it ?

> Today, we have items instead of themes. Do you also want to make that change?

Yes I think that this terminology is better.

> Also, having the list of background layers at the first level in the JSON object means that we have no association between themes and background layers. Right?

Yes ...

elemoine:
> This metadata is completely free word for c2cgeoportal, and only for Luxembourg, should we relay follow it ?

I think that 1) we should be consistent and use the same term everywhere, and 2) the name "background" is more explicit than the name "base".

sbrunner:
OK, than I will name it background_layers_group and background_layers.

elemoine:
Sounds good to me.

elemoine:
This is ready for review.

So, this PR:

* adds Angular UI Bootstrap (it is built as part of the project's build process)
* adds a backgroundlayer directive based on Angular UI Bootstrap's `dropdown` directive

With help from @pgiraud I fixed the CSS.

The thing that is still to be done is rely on the "themes" web service instead of hard-coding the list of background layers in the directive. But I suggest to do that with a separate PR, because more work on the "themes" web service (from @sbrunner) is needed.

Please review.

geoportailv3/static/js/backgroundlayer/backgroundlayer.html
```diff
@@ -0,0 +1,13 @@
+<div class="text-center">
+  <span translate>Background layer:&nbsp;</span>
```
pgiraud: We should probably put the `&nbsp;` outside of the translated string.

geoportailv3/static/js/backgroundlayer/backgroundlayer.html
```diff
@@ -0,0 +1,13 @@
...
+  <div class="dropdown" dropdown>
+    <button type="button" class="btn btn-default" dropdown-toggle>
+    {{ctrl.bgLayer.name}} <span class="caret"></span>
```
pgiraud: Missing indentation? I'm not sure.

geoportailv3/static/js/backgroundlayer/backgroundlayerdirective.js
```diff
@@ -0,0 +1,89 @@
+goog.provide('app.backgroundlayerDirective');
```
pgiraud: Missing fileoverview.

elemoine:
As discussed with @pgiraud this morning, I finally decided against using Angular UI Bootstrap for now. 

I tried porting the catalog directive to using Angular UI Bootstrap, but I failed. Using Angular UI Bootstrap's `accordion` directive in the catalog directive brought a lot of complexity. The `ngeo-layertree` is already quite complex, because of its recursive nature. Making `ngeo-layertree` and `accordion` play together turned out to be a nightmare. I could probably make it work, but it would take another day or so.

So the new version of the `app-backgroundlayer` directive is based on Bootstrap's `dropdown` jQuery plugin. The use of the plugin for this directive is clean.

I also took @pgiraud's good comments into account.

elemoine:
If there's no objection I will merge this when the Travis build passes again.

elemoine:
@sbrunner, I'd just like to know if you plan to work on the `themes` web service this week? (As discussed above) Thanks.

sbrunner:
I started working on this here: https://github.com/camptocamp/c2cgeoportal/tree/multi-theme :-)

PR #172, opened 2015-01-20
Panels for tools using ngeo-btn-group

pgiraud:
The goal of those changes is to prepare the work for the toolboxes (button with an attached panel which open when the button is clicked).
We use a ngeo-btn-group so that when a toolbox panel is shown the other toolboxes are hidden.

The same work needs to be done for the top right buttons (login, languages).

### DIFF
--- a/geoportailv3/static/js/maincontroller.js
+++ b/geoportailv3/static/js/maincontroller.js
@@ -46,12 +46,12 @@ app.MainController = function($scope, gettextCatalog, langUrlTemplate) {
   /**
    * @type {Boolean}
    */
-  this['sidebarOpen'] = false;
+  this['drawOpen'] = false;
 
   /**
    * @type {Boolean}
    */
-  this['mymapsOpen'] = false;
+  this['infosOpen'] = false;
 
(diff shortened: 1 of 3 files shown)


pgiraud:
Please review.

elemoine:
I'd just initialize `drawOpen` and the others in the main controller, as done for other state variables. Otherwise looks good to me.

geoportailv3/static/js/maincontroller.js
```diff
@@ -61,7 +61,27 @@ app.MainController = function($scope, gettextCatalog, langUrlTemplate) {
   /**
    * @type {Boolean}
    */
-  this['infosOpen'] = false;
+  this['measureOpen'] = false;
+
+  /**
+   * @type {Boolean}
```
elemoine: The correct spelling is `boolean` rather than `Boolean` I think, but that may be convention: lowercase for primitive types (`number`, `string`, `boolean`, `function`, `null`, `undefined`), and uppercase for objects (`Object`, `Array`...).

PR #174, opened 2015-01-21
Upgrade c2cgeoportal version

sbrunner:
Update to new version of c2cgeoportal,

One important change is that we merge the geoportailv3.pot and the geoportailv3-db.pot into one single file named geoportailv3-client.pot, => same for po files.


sbrunner:
Related to camptocamp/c2cgeoportal#1174

sbrunner:
The Travis error will be fixed when the c2cgeoportal will be merged :-)

elemoine:
Quick question: why did you move the less files into the `static` dir?

sbrunner:
Because all the client specific files are in the static folder static, for me it's relay strange to see those files here, where these is only files directly related to do the build or relay separate features like the the print, the deploy and mapserver ...

elemoine:
I understand. At the same time it is also strange and confusing to place in the `static` dir files that are not static resources (i.e served by the application as static assets). How about using `geoportailv3/less`?

sbrunner:
less files are dynamic files ?
why is it reals a goal to hide those files ?
and if one time we want to use a library like http://lesscss.org/#download-options to use the less files in the browser we should move them in the static folder ... ?
For those reason for me the best place is in the static folder...

petzlux:
Where did the db_to_pot helper function go ? I can see a call to a command db2po now, but not the function ?

sbrunner:
@petzlux to the c2cgeoportal project: https://github.com/camptocamp/c2cgeoportal/pull/1174/files#diff-67

petzlux:
Okay ! Thanks

elemoine:
I am not opposed to place the less files in the `static` folder.

petzlux:
Before this PR is merged, can we make sure that Transifex project will be updated, as we changed the .pot/.po file structure !

PR #179, opened 2015-01-22
added luxembourgish to locale names

petzlux:
added override of available_locale_names in geoportailv3 config, to add luxembourgish !

### DIFF
--- a/vars_geoportailv3.yaml
+++ b/vars_geoportailv3.yaml
@@ -25,6 +25,12 @@ vars:
     servers:
         my_maps: http://ws.geoportail.lu/mymaps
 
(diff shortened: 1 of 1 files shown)


sbrunner:
Curious to know where it is visible ?

petzlux:
languages variable is used in fulltextsearch in https://github.com/camptocamp/c2cgeoportal/search?utf8=%E2%9C%93&q=languages  . 

In our search, if we integrate content from CMS and also layer search, will need to know language to be searched for!?!

sbrunner:
ok, thanks, look good, merge :-)

PR #188, opened 2015-01-27
Translations for catalog and layermanager

pgiraud:
Please review.

### DIFF
--- a/geoportailv3/static/js/catalog/layertreenode.html
+++ b/geoportailv3/static/js/catalog/layertreenode.html
@@ -1,14 +1,14 @@
 <a data-toggle="collapse" data-parent="#catalog-{{::layertreenodeCtrl.parentUid}}"
    href="#catalog-{{::layertreenodeCtrl.uid}}"
    ng-if="::layertreenodeCtrl.node.children && layertreenodeCtrl.depth < 3">
-  {{::layertreenodeCtrl.node.name}}
+  {{::layertreenodeCtrl.node.name | translate}}
 </a>
 <span ng-if="::layertreenodeCtrl.node.children && layertreenodeCtrl.depth >= 3">
-  {{::layertreenodeCtrl.node.name}}
+  {{::layertreenodeCtrl.node.name | translate}}
 </span>
 <div ng-if="::!layertreenodeCtrl.node.children">
   <a class="btn btn-default" ng-click="catalogCtrl.getInfo(node, layer)">i</a>
-  {{::layertreenodeCtrl.node.name}}
+  {{::layertreenodeCtrl.node.name | translate}}
   <input type="checkbox" ng-if="::!layertreenodeCtrl.node.children"
       ng-model="layertreenodeCtrl.getSetActive" ng-model-options="{getterSetter: true}"/>
 </div>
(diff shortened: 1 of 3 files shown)


sbrunner:
camptocamp/c2cgeoportal#1197, camptocamp/pyramid_closure#44 will be a better solution

jaykayone:
There is a problem in the translations:
http://devv3.geoportail.lu/main/ vs http://emwelt.geoportail.lu/

In the old version there is a group called "env_bruit" which should be translated "bruit"... but it is translated "Aéroport 2011 (LNGT)"...

I guess it is a problem with the migration of the translations @petzlux ... in the old portal, the group was called "env_bruit" but it's label was "bruit" --> it's translation has not been migrated. we should check this. it does not affect this PR though ... I'll open an issue

geoportailv3/static/js/catalog/layertreenode.html
```diff
@@ -1,14 +1,14 @@
 <a data-toggle="collapse" data-parent="#catalog-{{::layertreenodeCtrl.parentUid}}"
    href="#catalog-{{::layertreenodeCtrl.uid}}"
    ng-if="::layertreenodeCtrl.node.children && layertreenodeCtrl.depth < 3">
-  {{::layertreenodeCtrl.node.name}}
+  {{::layertreenodeCtrl.node.name | translate}}
```
elemoine: Using the `translate` filter should be reserved to cases where using the `translate` directive is not possible. See https://angular-gettext.rocketeer.be/dev-guide/annotate/. In that case I think we can use the `translate` directive.

elemoine: And while we're at that we should remove the use of the `translate` filter in the index.html page: https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/templates/index.html#L57.

pgiraud: This doesn't seem to work unfortunately.

pgiraud: <translate>{{property}}</translate>

elemoine: That's stange. Not what I understand from https://angular-gettext.rocketeer.be/dev-guide/annotate/.

elemoine: It's probably me not understanding when the filter rather than the directive should be used.

elemoine: I better understand now. It is at least possible for the my_layers string, see https://github.com/Geoportail-Luxembourg/geoportailv3/commit/1936a35883b0eb323de09b03b3839bc5c9f6dc64. And I did the test: using the directive rather than the filter removes 3 watchers.

elemoine: @pgiraud if you agree with https://github.com/Geoportail-Luxembourg/geoportailv3/commit/1936a35883b0eb323de09b03b3839bc5c9f6dc64 please add it to your branch before merging. Thanks.

pgiraud: done

PR #191, opened 2015-01-27
Transifex Client and resource fixes - i18n

petzlux:


sbrunner:
What does it will fix ?

sbrunner:
It just seem to duplicate things ...

petzlux:
@sbrunner okay, I understand now, looking at makefile, geoportailv3-db.pot & -js.pot are merged to  geoportailv3-client.pot !

Shall we add -db.pot and -js.pot to .gitignore, or add both to the project ?

sbrunner:
Good question, 
Traditionally we commit them, but I think that not needed ...
@elemoine what do you think ?

petzlux:
Okay changed branch to reflect comments, waiting on decision if we exclude or include -db.pot & -js.pot

elemoine:
+1 for not committing them into the project. But in that case I think they should be generated in the `.build` directory, instead of in the `$(PACKAGE)/locale` directory.

sbrunner:
I just create a c2cgeoportal issue for that :-) https://github.com/camptocamp/c2cgeoportal/issues/1199

PR #193, opened 2015-01-27
Adding a make target to automatically build less files

pgiraud:


pgiraud:
This pull request doesn't have any influence on the project itself. Merging.

elemoine:
Is `watchmedo` better than `nosier`?

pgiraud:
For an unknown reason, nosier refused to work properly today. I decided to find a less old package to do the same.

elemoine:
Ok thanks.

PR #194, opened 2015-01-27
No capital case for boolean type def

pgiraud:


elemoine:
Please merge when Travis is happy.

Issue #200, opened 2015-01-29
Modified .po files in the working dir [c2c]


elemoine:
Restant c2c 01/02/2015 1j

----

With the latest `master` I get modified `.po` and `.pot`  files (reported by Git as such) after a build. Is this related to the discussion in https://github.com/Geoportail-Luxembourg/geoportailv3/pull/191? Is someone going to work on this?

geoportallux:
patrick is away until monday...
maybe sbrunner?

I remember that the files were included in a commit for travis to be happy ...
But I dont know any details


Jeff Konnen

Chargé d’études – informaticien/Chef de projet



LE GOUVERNEMENT DU GRAND-DUCHÉ DE LUXEMBOURG

Ministère des Finances

Administration du cadastre et de la topographie

Service informatique & Service géoportail



54, avenue Gaston Diderich<x-apple-data-detectors://1/0> . L-1420 Luxembourg

Tél. (+352) 44901-261<tel:(+352)%252044901-261> . Fax (+352) 44901-333<tel:(+352)%252044901-333>

E-mail: jeff.konnen@act.etat.lu<mailto:jeff.konnen@act.etat.lu>

www.act.public.lu<http://www.act.public.lu/> www.geoportail.lu<http://www.geoportail.lu/>

On 29.01.2015, at 09:11, Éric Lemoine <notifications@github.com<mailto:notifications@github.com>> wrote:


With the latest master I get modified .po and .pot files (reported by Git as such) after a build. Is this related to the discussion in #191<https://github.com/Geoportail-Luxembourg/geoportailv3/pull/191>? Is someone going to work on this?

sbrunner:
I don't have any pot modified, for the po at each build he ge the new files from transifex if we don't want that we should remove this line: https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3.mk#L19

petzlux:
+1 for excluding transifex push pull from build

elemoine:
@sbrunner I have `geoportailv3/locale/geoportailv3-db.pot`.

sbrunner:
* [ ] Transifex shouldn't be called every times
* [ ] All pot should be removed from the repository
* [ ] All pot should be in the git ignore

PR #234, opened 2015-02-02
Update translations fix

petzlux:
removed the build target update-translations as a pre-rule, to speed up time to serve. If push/pull from transifex is now needed, the build target has to be explicitely called.  also ignores now -db and -js pot files, as they are only intermediary steps in the build process. 

@sbrunner afaik removing the pre rule doesn't impact the build process negatively, can you confirm ?


petzlux:
should also help with issue #200

sbrunner:
Yes that right :-)

PR #238, opened 2015-02-02
Layerinfo directive & service

pgiraud:
For now, it doesn't do much but displays an almost empty popup when clicking on the "info" button in both catalog and layer manager.
It should eventually call a service to get metadata info.
Please review.

### DIFF
--- a/geoportailv3/static/js/catalog/layertreenode.html
+++ b/geoportailv3/static/js/catalog/layertreenode.html
@@ -7,7 +7,7 @@
   {{::layertreenodeCtrl.node.name | translate}}
 </span>
 <div ng-if="::!layertreenodeCtrl.node.children">
-  <a class="btn btn-default" ng-click="catalogCtrl.getInfo(node, layer)">i</a>
+  <app-layerinfo app-layerinfo-layer="::layertreenodeCtrl.layer"></app-layerinfo>
   {{::layertreenodeCtrl.node.name | translate}}
   <input type="checkbox" ng-if="::!layertreenodeCtrl.node.children"
       ng-model="layertreenodeCtrl.getSetActive" ng-model-options="{getterSetter: true}"/>
--- a/geoportailv3/static/js/layerinfo/layerinfodirective.js
+++ b/geoportailv3/static/js/layerinfo/layerinfodirective.js
@@ -1,0 +1,49 @@
+goog.provide('app.layerinfoDirective');
+
+goog.require('app');
+goog.require('app.ShowLayerinfo');
+
+
+/**
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+app.layerinfoDirective = function(appLayerinfoTemplateUrl) {
+  return {
+    restrict: 'E',
+    scope: {
+      'layer': '=appLayerinfoLayer'
+    },
+    controller: 'AppLayerinfoController',
+    controllerAs: 'ctrl',
+    bindToController: true,
+    templateUrl: appLayerinfoTemplateUrl
+  };
+};
+
+app.module.directive('appLayerinfo', app.layerinfoDirective);
+
+
+/**
+ * @constructor
+ * @param {app.ShowLayerinfo} appShowLayerinfo app.ShowLayerinfo service.
+ * @ngInject
+ * @export
+ */
+app.LayerinfoController = function(appShowLayerinfo) {
+  /**
+   * @private
+   */
+  this.showLayerInfo_ = appShowLayerinfo;
+};
+
+
+/**
+ * @export
+ */
+app.LayerinfoController.prototype.getInfo = function() {
+  this.showLayerInfo_(this['layer']);
+};
+
+
+app.module.controller('AppLayerinfoController', app.LayerinfoController);
--- a/geoportailv3/static/js/layerinfo/layerinfoservice.js
+++ b/geoportailv3/static/js/layerinfo/layerinfoservice.js
@@ -1,0 +1,43 @@
+/**
+ * @fileoverview Provides a layer info service. That service is a function used
+ * to retrieve and display the info (metadata) for a layer.
+ */
+
+goog.provide('app.ShowLayerinfo');
+
+goog.require('ngeo.CreatePopup');
+
+/**
+ * @typedef {function(ol.layer.Layer)}
+ */
+app.ShowLayerinfo;
+
+
+/**
+ * @param {angular.$http} $http Angular $http service
+ * @param {angular.$sce} $sce Angular $sce service
+ * @param {ngeo.CreatePopup} ngeoCreatePopup Ngeo popup factory service
+ * @ngInject
+ */
+app.showLayerinfoFactory = function($http, $sce, ngeoCreatePopup) {
+
+  /**
+   * @type {ngeo.Popup}
+   */
+  var popup = ngeoCreatePopup();
+
+  return (
+
+      /**
+       * @param {ol.layer.Layer} layer The layer
+       */
+      function(layer) {
+        var title = /** @type {string} */ (layer.get('label'));
+        popup.setTitle(title);
+        var content = $sce.trustAsHtml("<b>some</b> content");
+        popup.setContent(content);
+        popup.show();
+      });
+};
+
+app.module.service('appShowLayerinfo', app.showLayerinfoFactory);
--- a/geoportailv3/static/js/layerinfo/popup.html
+++ b/geoportailv3/static/js/layerinfo/popup.html
@@ -1,0 +1,5 @@
+<h4 class="popover-title ngeo-popup-title">
+  <span ng-bind="title | translate"></span>
+  <button type="button" class="close" ng-click="open = false"> &times;</button>
+</h4>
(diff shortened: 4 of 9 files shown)


elemoine:
It looks like this needs a rebase.

geoportailv3/static/js/catalog/layertreenode.html
```diff
@@ -7,7 +7,7 @@
   {{::layertreenodeCtrl.node.name | translate}}
 </span>
 <div ng-if="::!layertreenodeCtrl.node.children">
-  <a class="btn btn-default" ng-click="catalogCtrl.getInfo(node, layer)">i</a>
+  <app-layerinfo app-layer="layertreenodeCtrl.layer"></app-layerinfo>
```
elemoine: One-time binding for `layertreenodeCtrl.layer`?

geoportailv3/static/js/layerinfo/layerinfodirective.js
```diff
@@ -0,0 +1,51 @@
...
+
+
+/**
+ * @constructor
+ * @param {app.ShowLayerinfo} appShowLayerinfo app.ShowLayerinfo service.
+ * @ngInject
+ * @export
+ */
+app.LayerinfoController = function(appShowLayerinfo) {
+  this['uid'] = goog.getUid(this);
```
elemoine: Is this used?

geoportailv3/static/js/layerinfo/layerinfoservice.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * @fileoverview Provides a layer info service. That service is a function used
+ * to retrieve the info (metadata) for a layer.
```
elemoine: "to retrieve" -> "to retrieve and display"

geoportailv3/static/js/layerinfo/popup.html
```diff
@@ -0,0 +1,5 @@
+<h4 class="popover-title ngeo-popup-title">
+  <span>{{title | translate}}</span>
```
elemoine: Does it work with the following:

```html
  <span ng-bind="title | translate"></span>
```

Just curious.

geoportailv3/static/js/layerinfo/popup.html
```diff
@@ -0,0 +1,5 @@
...
+  <button type="button" class="close" ng-click="open = false"> &times;</button>
+</h4>
+<div class="popover-content" ng-bind-html="content"></div>
```
elemoine: Do we need to translate the content? Or does the metadata service return already-translated responses? @jaykayone, any idea?

geoportailv3/static/js/layerinfo/layerinfodirective.js
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+app.layerinfoDirective = function(appLayerinfoTemplateUrl) {
+  return {
+    restrict: 'E',
+    scope: {
+      'layer': '=appLayer'
```
elemoine: Our convention is to repeat the name of the directive, so `appLayerinfoLayer` instead of `appLayer`. I know it's more verbose, but I'd favor consistency.

geoportailv3/static/js/layerinfo/popup.html
pgiraud: It does.

geoportailv3/static/js/layerinfo/layerinfodirective.js
pgiraud: Nope. Removing it.

geoportailv3/static/js/layerinfo/layerinfoservice.js
```diff
@@ -0,0 +1,44 @@
...
+      /**
+       * @param {ol.layer.Layer} layer The layer
+       */
+      function(layer) {
+        popup.show();
+        var title = /** @type {string} */ (layer.get('label'));
+        popup.setTitle(title);
+        var content = /** @type {string} */
+            ($sce.trustAsHtml("<b>some</b> content"));
+        popup.setContent(content);
```
elemoine: Would it make more sense to set the title and the content and **then** show the popup?

geoportailv3/static/js/layerinfo/popup.html
jaykayone: there are two sides to this.
it will be presented as a table with titles in the first td and content in the second td.

The first column has to be translated, the content comes directly translated, but the language has to included in the call

eg: http://shop.geoportail.lu/Portail/inspire/webservices/getMD.jsp?uid=ae1bc422-5751-4efa-92af-c24287609dbb&lang=fr&_dc=1422894522958&cb=stcCallback1002

geoportailv3/static/js/layerinfo/layerinfoservice.js
```diff
@@ -0,0 +1,44 @@
...
+
+      /**
+       * @param {ol.layer.Layer} layer The layer
+       */
+      function(layer) {
+        popup.show();
+        var title = /** @type {string} */ (layer.get('label'));
+        popup.setTitle(title);
+        var content = /** @type {string} */
+            ($sce.trustAsHtml("<b>some</b> content"));
```
elemoine: It is strange that `trustAsHtml`'s return type is `*` rather than `string`. Any idea why?

elemoine:
I've added a couple comments, otherwise looks very good to me.

geoportailv3/static/js/layerinfo/layerinfoservice.js
pgiraud: It would make sense for now. But when we have a working service, I think will be nice to show a popup displaying a loading message while waiting for the request response.

elemoine: Right.

geoportailv3/static/js/layermanager/layermanager.html
```diff
@@ -1,6 +1,7 @@
 <div ng-if="!ctrl.layers.length" translate>No layer selected</div>
 <ul id="layermanager-{{::ctrl.uid}}" class="layermanager">
   <li ng-repeat="layer in ctrl.layers" class="panel">
+    <app-layerinfo app-layer="layer"></app-layerinfo>
```
elemoine: One-time binding?

geoportailv3/static/js/layermanager/layermanager.html
```diff
@@ -1,6 +1,7 @@
 <div ng-if="!ctrl.layers.length" translate>No layer selected</div>
 <ul id="layermanager-{{::ctrl.uid}}" class="layermanager">
   <li ng-repeat="layer in ctrl.layers" class="panel">
+    <app-layerinfo app-layerinfo-layer="layer"></app-layerinfo>
```
elemoine: One-time binding again here?

(Repeating this comment because of the rebase.)

geoportailv3/static/js/layerinfo/layerinfoservice.js
pgiraud: No. It's a bit oscure to me.

elemoine: On Mon, Feb 2, 2015 at 5:43 PM, Pierre GIRAUD <notifications@github.com> wrote:
>
> In geoportailv3/static/js/layerinfo/layerinfoservice.js:
>
> > +  /**
> > +   * @type {ngeo.Popup}
> > +   */
> > +  var popup = ngeoCreatePopup();
> > +
> > +  return (
> > +
> > +      /**
> > +       * @param {ol.layer.Layer} layer The layer
> > +       */
> > +      function(layer) {
> > +        popup.show();
> > +        var title = /** @type {string} */ (layer.get('label'));
> > +        popup.setTitle(title);
> > +        var content = /** @type {string} */
> > +            ($sce.trustAsHtml("<b>some</b> content"));
>
> No. It's a bit oscure to me.


Actually, $sce.trustAsHtml does not return a string. It returns an
object that can be used with ng-bind-html. So casting to a string does
not really make sense. The problem is in ngeo, in the signature of the
setContent function.



-- 
Eric Lemoine

Camptocamp France SAS
Savoie Technolac, BP 352

Tel : 00 33 4 79 44 44 94
Mail : eric.lemoine@camptocamp.com
http://www.camptocamp.com

geoportailv3/static/js/layerinfo/layerinfoservice.js
```diff
@@ -0,0 +1,44 @@
...
+  return (
+
+      /**
+       * @param {ol.layer.Layer} layer The layer
+       */
+      function(layer) {
+        var title = /** @type {string} */ (layer.get('label'));
+        popup.setTitle(title);
+        var content = /** @type {string} */
+            ($sce.trustAsHtml("<b>some</b> content"));
```
elemoine: We can remove that typecast when https://github.com/camptocamp/ngeo/pull/157 is merged.

elemoine:
Jeff's comment to take into account: https://github.com/Geoportail-Luxembourg/geoportailv3/pull/238#discussion_r23936032.

PR #240, opened 2015-02-03
Map is 100% in full screen mode

### LINKED ISSUE #197: Bad interaction between Fullscreen and left panel [lux]
If the left panel is open and if we enter in the fullscreen mode, the map is displayed between 2 black boxes.
Modifying the property width of the map from "auto" to "100%" solves the problem. But the left panel does not work as expected anymore.

pgiraud:
Fixes #197.
Please review.


rmichaelis:
Looks good to me.

Issue #243, opened 2015-02-04
Language Switcher UI adjustments


petzlux:
Gap between tiles is too large, and also slight overlap on top with header bar

elemoine:
FWIW, I agree.

jaykayone:
In the design, I also saw that the color is theme-dependant
![screen shot 2015-02-04 at 11 48 54](https://cloud.githubusercontent.com/assets/1518019/6039125/d7d22b20-ac63-11e4-9013-5cce09e7cabf.png)

PR #244, opened 2015-02-04
language switcher styling

petzlux:
partially adresses #243 . 

@pgiraud I cant seem to be able to fix gaps between tiles for language switcher menu , can you help ?


pgiraud:
This was a bit tricky.
Waiting for travis tests to pass and then merging.

PR #245, opened 2015-02-05
 Remove mainCtrl.sidebarOpen

elemoine:
This PR removes the `mainCtrl.sidebarOpen` state and the four watchers that are necessary to keep it in sync with `layersOpen`, `mymapsOpen` and `infosOpen`.

With this PR we go from 243 to 239 watchers, 208 of which coming from the catalog directive (with the per-node watchers we need).


pgiraud:
Really nice improvement.
Please merge once travis is happy.

PR #248, opened 2015-02-05
Removed unrequired translate

pgiraud:
We don't need to translate the string in the language switcher.
Please review.


geoportailv3/templates/index.html
```diff
@@ -32,16 +32,16 @@
           <li class="lang icon hidden-xs" ng-class="mainCtrl.languageOpen ? 'active': ''">
             <a href translate ngeo-btn ng-model="mainCtrl.languageOpen">lang</a>
             <ul class="nav toolbox-panel toolbox-panel-top">
-              <li translate ng-class="mainCtrl.lang == 'lb' ? 'active' : ''">
```
elemoine: While you're at it, I think `"{active: mainCtrl.lang == 'lb'}"` offers a nicer syntax in this case.

elemoine:
I've just added a comment suggesting to use another syntax for ng-class in that case. In any way, please merge.

PR #251, opened 2015-02-09
authentication

rmichaelis:
Please review. 

1) Some new parameters have to be added in config such as ldap_url ldap_bind ldap_passwd ldap_base_dn
Up to now these parameters are in my vars_renaud_yaml. I'm not sure if I have to add it anywhere else.
If these parameters are not set, Travis failed. 

2) library for openldap-devel has to be installed.

### DIFF
--- a/geoportailv3.mk
+++ b/geoportailv3.mk
@@ -16,6 +16,8 @@ PRINT3 = FALSE
 
 DISABLE_BUILD_RULES = test-packages test-packages-ngeo
 
+CONFIG_VARS += ldap_url ldap_bind ldap_passwd ldap_base_dn
+
 include CONST_Makefile
 
 DEV_REQUIREMENTS += git+https://github.com/transifex/transifex-client.git@fix-proxies#egg=transifex-client-proxies
--- a/geoportailv3.mk
+++ b/geoportailv3.mk
@@ -16,7 +16,7 @@ PRINT3 = FALSE
 
 DISABLE_BUILD_RULES = test-packages test-packages-ngeo
 
-CONFIG_VARS += ldap_url ldap_bind ldap_passwd ldap_base_dn
+CONFIG_VARS += ldap
 
 include CONST_Makefile
 
--- a/geoportailv3/static/js/usercontroller.js
+++ b/geoportailv3/static/js/usercontroller.js
@@ -69,7 +69,7 @@ app.UserController = function($http) {
 
 
 /**
- * @param {object} credentials Credentials.
+ * @param {Object} credentials Credentials.
  * @export
  */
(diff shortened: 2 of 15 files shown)


geoportailv3.mk
```diff
@@ -16,6 +16,8 @@ PRINT3 = FALSE
 
 DISABLE_BUILD_RULES = test-packages test-packages-ngeo
 
+CONFIG_VARS += ldap_url ldap_bind ldap_passwd ldap_base_dn
```
sbrunner: Those variables are missing in the `geoportailv3_vars.yaml` file ...

Not important but I prefer to organized them like it:
```
ldap:
    url: ...
    bind: ...
    passwd: ...
    base_dn: ...
```
than the above line become `CONFIG_VARS += ldap`
and the init:
```
ldap_settings = onfig.get_settings()["ldap"]
config.ldap_setup(
  ldap_settings['url'],
  ldap_settings['bind'],
  ldap_settings['passwd'],
)

config.ldap_set_login_query(
  ldap_settings['base_dn'],
  ...
```
Not a heavy change just more organized settings :-)

rmichaelis:
Please review.

sbrunner:
For me that looks good :-)

geoportailv3/static/js/usercontroller.js
```diff
@@ -0,0 +1,191 @@
...
+  this['credentials'] = {
+    'login' : null,
+    'password' : null
+  };
+
+  /**
+   * @type {angular.$http}
+   * @private
+   */
+  this['http_'] = $http;
```
elemoine: You don't need/want to use the bracket notation for `http_` because that property is not used in the view. Using the bracket notation and quotes prevent renaming. We want to use it only when necessary.

geoportailv3/templates/index.html
```diff
@@ -28,9 +28,46 @@
 
         <ul class="nav navbar-nav pull-right">
           <li class="visible-xs-inline-block search icon"><a href translate>search</a></li>
-          <li class="user icon"><a href translate>user</a></li>
+          <li class="user icon" ng-class="mainCtrl.userOpen ? 'active': ''">
+            <a href translate ngeo-btn ng-model="mainCtrl.userOpen" ng-click="mainCtrl.languageOpen = false">user</a>
+            <div ng-controller="UserController as userCtrl" class="navigation-account user-panel toolbox-panel toolbox-panel-top second-level">
```
elemoine: As discussed in the ngeo guidelines and during the workshop we want to limit the use of `ng-controller`. So, instead of using `ng-controller`, I would have created an `app-authentication` directive, with its own `AppAuthenticationController`, as done for other parts of the application.

PR #259, opened 2015-02-17
Measure tools

pgiraud:
Adds tools to measure length, area, and azimut.

### DIFF
--- a/build.json
+++ b/build.json
@@ -34,6 +34,7 @@
     "js": [
       "geoportailv3/static/js/main.js",
       "node_modules/openlayers/externs/olx.js",
+      "node_modules/ngeo/externs/ngeox.js",
       "node_modules/openlayers/externs/oli.js"
     ],
     "jscomp_error": [
--- a/geoportailv3/static/js/main.js
+++ b/geoportailv3/static/js/main.js
@@ -10,12 +10,12 @@
 goog.provide('app_main');
 
 goog.require('app.MainController');
-goog.require('app.MeasureController');
 goog.require('app.authenticationDirective');
 goog.require('app.backgroundlayerDirective');
 goog.require('app.catalogDirective');
 goog.require('app.layerinfoDirective');
 goog.require('app.layermanagerDirective');
+goog.require('app.measureDirective');
 goog.require('app.scalelineDirective');
 goog.require('ngeo.btngroupDirective');
 goog.require('ngeo.resizemapDirective');
--- a/geoportailv3/static/js/main.js
+++ b/geoportailv3/static/js/main.js
@@ -10,6 +10,7 @@
 goog.provide('app_main');
 
 goog.require('app.MainController');
+goog.require('app.MeasureController');
 goog.require('app.authenticationDirective');
 goog.require('app.backgroundlayerDirective');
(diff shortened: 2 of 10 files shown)


pgiraud:
Hi all, now that the measure interactions have been added to ngeo, the current pull request is ready for review.
Can someone on Luxembourg's side give it a try? Don't forget to update the node modules.

jaykayone:
OK I'll give it a try

jaykayone:
This is truly great work.

Things I miss:
- icons
- surface fill color
- for length and surface the moment where you go from m to km resp m2 to km2 is too early. For length, I would only switch to km at 1000m (and not 10m), for surfaces, I would also use ar and hectar
- Azimut: Distance and elevation offset in the label

But I guess all these points can be done by @rmichaelis 

some minor "bugs" we saw:
- when you zoom out while measuring a surface, the label becomes too big and you cannot place a vertex on / under the label
- the azimut tool blocks panning. you cannot pan unless you deactivate the tool
- when clicking X to close the measure panel, tools are not deactivated

pgiraud:
> icons

Can be done on your side.

> surface fill color

Can be configured in the application.

> for length and surface the moment where you go from m to km resp m2 to km2 is too early. For length, I would only switch to km at 1000m (and not 10m), for surfaces, I would also use ar and hectar

I think we need a better algorithm with number of digits precision.
For surfaces in ares and hectares, I propose we address this later.

> Azimut: Distance and elevation offset in the label

Distance in the tooltip can be easily added.
Elevation offset will be a bit trickier. Can you please open an issue for this?

> when you zoom out while measuring a surface, the label becomes too big and you cannot place a vertex on / under the label

Do someone really want to draw a polygon in such a context?

> the azimut tool blocks panning. you cannot pan unless you deactivate the tool

See https://github.com/camptocamp/ngeo/pull/166.

> when clicking X to close the measure panel, tools are not deactivated

This should be addressed in the application in my opinion.

jaykayone:
OK, we'll wait for you to merge this PR so @rmichaelis can take over

pgiraud:
up

elemoine:
Looking at it now…

elemoine:
I suggest to not use `ng-controller` and use a measure directive instead. See https://github.com/Geoportail-Luxembourg/geoportailv3/compare/elemoine_measure.

jaykayone:
I have just seen a big problem.
The measures are taken in EPSG:3857, whereas we need to have them in EPSG:2169.

This makes a huge difference. A distance measured as 200m in webmercator makes 130m in LUREF!

elemoine:
http://blogs.esri.com/esri/arcgis/2010/03/05/measuring-distances-and-areas-when-your-map-uses-the-mercator-projection/ is interesting.

jaykayone:
I see that your googling skills are the same than ours :-)

elemoine:
@jaykayone @pgiraud and I discussed it and we think there are solutions. So I'd suggest that we merge this and address the remaining issues with separate PRs.

jaykayone:
ok, can you open a separate issue for this please?

elemoine:
@pgiraud, please rebase and merge. I'm going to create an issue for the projection issue.

PR #271, opened 2015-02-24
Set PRINT_OUTPUT to /tmp in elemoine.mk

elemoine:


pgiraud:
Can you please do the same in my config file? I'll need this too.

elemoine:
> Can you please do the same in my config file? I'll need this too.

Done.

pgiraud:
Please merge, when happy.

sbrunner:
If should also be possible to add the following line:
`DEFAULT_BUILD_RULES = print`
:-)

elemoine:
> If should also be possible to add the following line:
> DEFAULT_BUILD_RULES = print
> :-)

What does that mean exactly?

sbrunner:
The CONST_Makefile build a set of rules that should be run by the build rule, and with DEFAULT_BUILD_RULES we can remove element in the list :-)

elemoine:
> The CONST_Makefile build a set of rules that should be run by the build rule, and with DEFAULT_BUILD_RULES we can remove element in the list :-)

Ok, thanks.

Issue #201, opened 2015-01-29
angular-animate.min.js.map 404 (Not found) error in non-debug mode


elemoine:
We get this error at application load time in non-debug mode:

```
GET http://localhost:5000/proj/build/angular-animate.min.js.map 404 (Not Found) 
```

This is annoying and should be fixed eventually.

jaykayone:
who's in charge?

elemoine:
I can look into it.

PR #274, opened 2015-02-24
Remove sourceMappingURL from build.js

elemoine:
This is to prevent 404s on source map files in non-debug mode.

Fixes https://github.com/Geoportail-Luxembourg/geoportailv3/issues/201.


sbrunner:
no problem for me :-)

PR #276, opened 2015-02-24
Add an appThemes service

elemoine:
This PR adds an `appThemes` service to the project's Angular application. This service requests the "themes" data from c2cgeoportal's `themes` web service, and provides functions to interact with that "themes" data. The service is used by the catalog and background layer directives.

Now that the background layer directive gets the list of background layers from the "themes" web service, we have 5 background layers in the background layer selector. BUT only two layers actually work: `basemap_global` and `topogr_global`. I would suggest that we look at the non-working background layers after merging this.

Please review.

### DIFF
--- a/geoportailv3/static/js/themesservice.js
+++ b/geoportailv3/static/js/themesservice.js
@@ -1,0 +1,115 @@
+/**
+ * @fileoverview This file defines the Themes service. This service interacts
+ * with c2cgeoportal's "themes" web service and exposes functions that return
+ * objects in the tree returned by the "themes" web service.
+ */
+goog.provide('app.Themes');
+
+goog.require('app');
+goog.require('goog.asserts');
+
+
+/**
+ * @typedef {{themes: Array.<Object>, background_layers: Array.<Object>}}
+ */
+app.ThemesResponse;
+
+
+
+/**
+ * @constructor
+ * @param {angular.$http} $http Angular http service.
+ * @param {string} treeUrl URL to "themes" web service.
+ */
+app.Themes = function($http, treeUrl) {
+  /**
+   * @type {angular.$q.Promise}
+   * @private
+   */
+  this.promise_ = $http.get(treeUrl).then(
+      /**
+       * @param {angular.$http.Response} resp Ajax response.
+       * @return {Object} The "themes" web service response.
+       */
+      function(resp) {
+        return /** @type {app.ThemesResponse} */ (resp.data);
+      });
+};
+
+
+/**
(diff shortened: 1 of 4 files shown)


jaykayone:
Mea culpa for the not-working-layers.
I am aware of this. Sorry.
Shouldn't have any influence on the code

geoportailv3/static/js/themesservice.js
```diff
@@ -0,0 +1,118 @@
...
+ */
+app.ThemesResponse;
+
+
+
+/**
+ * @constructor
+ * @param {angular.$http} $http Angular http service.
+ * @param {string} treeUrl URL to "themes" web service.
+ */
```
pgiraud: don't we need ngInject here?

geoportailv3/static/js/themesservice.js
```diff
@@ -0,0 +1,118 @@
...
+
+
+/**
+ * Find an object by its name.
+ * @param {Array.<Object>} objects Array of objects.
+ * @param {string} objectName The object name.
+ * @return {Object} The object.
+ * @private
+ */
+app.Themes.findObject_ = function(objects, objectName) {
```
pgiraud: This method should actually be called findObjectByName.

geoportailv3/static/js/themesservice.js
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Find a theme object by its name.
+ * @param {Array.<Object>} themes Array of "theme" objects.
+ * @param {string} themeName The theme name.
+ * @return {Object} The theme object.
+ * @private
+ */
+app.Themes.findTheme_ = function(themes, themeName) {
+  return app.Themes.findObject_(themes, themeName);
```
pgiraud: What about using something like
goog.array.find(themes, function(theme, index) {return theme['name'] === objectName;});

geoportailv3/static/js/themesservice.js
elemoine: No, the services are injected in the factory function `app.ThemesFactory_`.

geoportailv3/static/js/themesservice.js
elemoine: Ok, no problem, I can change the name.

geoportailv3/static/js/themesservice.js
elemoine: Yep, I can use that.

elemoine:
@pgiraud, I changed the code as requested. I'll merge when Travis is happy.

PR #277, opened 2015-02-24
Fix the ngeo debug mode with Cache Buster

sbrunner:
requires: https://github.com/camptocamp/pyramid_closure/pull/49

### DIFF
--- a/CONST_vars.yaml
+++ b/CONST_vars.yaml
@@ -79,21 +79,18 @@ vars:
         config: "{directory}/jsbuild/app.cfg"
         root_dir: "{directory}"
 
+    # used for the "node_modules" and "closure" static views
+    node_modules_path: "{package}:node_modules"
     # pyramid_closure configuration
-    closure_library_path: 'process.stdout.write(require("openlayers/node_modules/closure-util").getLibraryPath())'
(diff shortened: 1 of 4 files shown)


CONST_vars.yaml
```diff
@@ -82,12 +82,15 @@ vars:
...
         - "{closure_library_path}/closure/goog"
         roots_with_prefix:
-            ../../../../../../../../../proj/js: "{directory}/{package}/static/js"
-            ../../../../../../../../../node_modules/openlayers: "{directory}/node_modules/openlayers"
-            ../../../../../../../../../node_modules/ngeo/src: "{directory}/node_modules/ngeo/src"
+            geoportailv3:static/js:
+                root: "{directory}/{package}/static/js"
+                is_pyramid_static_view: true
+            /{instanceid}/wsgi/node_modules/openlayers: "{directory}/node_modules/openlayers"
+            /{instanceid}/wsgi/node_modules/ngeo/src: "{directory}/node_modules/ngeo/src"
```
elemoine: I guess this won't work with `pserve`, which is what we use for development.

elemoine:
So, the current problem is that requests to `/proj` static resources yield 404's. I think the problem is related to the fact that we use the cache buster for `proj` but not for `closure`. What is the reason for not using the cache buster for `closure` and `node_modules`?

elemoine:
Superseded by https://github.com/Geoportail-Luxembourg/geoportailv3/pull/279.

PR #279, opened 2015-02-26
Change pyramid_closure config for camptocamp/pyramid_closure#50

elemoine:
This PR aims to fix the javascript debug mode that was broken by the update to the latest c2cgeoportal version.

Depends on camptocamp/pyramid_closure#50, which should be merged first.

### DIFF
--- a/CONST_vars.yaml
+++ b/CONST_vars.yaml
@@ -82,12 +82,11 @@ vars:
     # pyramid_closure configuration
     closure_library_path: 'process.stdout.write(require("openlayers/node_modules/closure-util").getLibraryPath())'
     pyramid_closure:
-        roots:
-        - "{closure_library_path}/closure/goog"
         roots_with_prefix:
-            ../../../../../../../../../proj/js: "{directory}/{package}/static/js"
-            ../../../../../../../../../node_modules/openlayers: "{directory}/node_modules/openlayers"
-            ../../../../../../../../../node_modules/ngeo/src: "{directory}/node_modules/ngeo/src"
+        - ["{closure_library_path}/closure/goog", "{closure_library_path}/closure/goog"]
+        - ["{package}:static/js", "{directory}/{package}/static/js"]
+        - ["{node_modules_path}/openlayers", "{node_modules_path}/openlayers"]
+        - ["{node_modules_path}/ngeo/src", "{node_modules_path}/ngeo/src"]
 
     # used for the "node_modules" and "closure" static views
     node_modules_path: "{directory}/node_modules"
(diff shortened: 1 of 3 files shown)


elemoine:
I've discovered a new problem related to the cache buster. When the cache buster is enabled Pyramid attempts to read the file that is passed to `request.static_url`. But in the case of the [`geoportailv3.json` lang file](https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/templates/index.html#L205) we pass to `static_url` a path that does not exist on the file system (because of `__lang__` pattern). I need to find a solution to this problem as well.

elemoine:
> I've discovered a new problem related to the cache buster. When the cache buster is enabled Pyramid attempts to read the file that is passed to request.static_url. But in the case of the geoportailv3.json lang file we pass to static_url a path that does not exist on the file system (because of __lang__ pattern). I need to find a solution to this problem as well.

I just added https://github.com/Geoportail-Luxembourg/geoportailv3/commit/c9ac120fa0f218199261854dec8a05ea23ed34f3 for that problem.

CONST_vars.yaml
```diff
@@ -82,12 +82,11 @@ vars:
...
-        roots:
-        - "{closure_library_path}/closure/goog"
         roots_with_prefix:
-            ../../../../../../../../../proj/js: "{directory}/{package}/static/js"
-            ../../../../../../../../../node_modules/openlayers: "{directory}/node_modules/openlayers"
-            ../../../../../../../../../node_modules/ngeo/src: "{directory}/node_modules/ngeo/src"
+        - "{closure_library_path}/closure/goog {closure_library_path}/closure/goog"
+        - "{package}:static/js {directory}/{package}/static/js"
+        - "{node_modules_path}/openlayers {node_modules_path}/openlayers"
+        - "{node_modules_path}/ngeo/src {node_modules_path}/ngeo/src"
```
sbrunner: We speak about having a config like `- ["{node_modules_path}/ngeo/src", "{node_modules_path}/ngeo/src"]` to don't post process the strings ...

elemoine:
@sbrunner, do you like it best?

sbrunner:
Yes I think that relay better :-)

elemoine:
This change requires having the latest version of `pyramid_closure`. To install the latest version of `pyramid_closure` you can use the following:

```shell
$ rm .build/requirements.timestamp
$ make -f elemoine.mk .build/requirements.timestamp
```

@sbrunner, maybe there's a better/simpler way?

sbrunner:
A simplest way it to do a `clean`/`build` but I prefer your solution, or eventually:
```
rm .build/requirements.timestamp
make -f elemoine.mk build
```

elemoine:
Thank you Stéphane.

PR #282, opened 2015-02-26
print parameter

rmichaelis:


renaud.mk
```diff
@@ -1,4 +1,4 @@
 INSTANCE_ID = renaud
 VARS_FILE = vars_renaud.yaml
-
+PRINT_OUTPUT = /tmp
```
jaykayone: shouldn't we include that param in geoportail_v3.mk rather than adding it in every make file?
in main.mk we can still oberride it with another dir if necessary

rmichaelis: +1

elemoine: Note that using `/tmp` does not work as initially expected, because it leads the Makefile rule to do a mv command with the dest file being the same as the source file, which produces an error. We need to find another way.

sbrunner: Shouldn't we create a new file like `luxambourg_demo.mk` (and `camptocamp_demo.mk` ?) where we can put the host specific things ?

jaykayone: yes, but they don't need to be in git...
I'd like to prevent that everyone sets a certain variable to the same value in his make file that's why I suggested to put it in geoportailv3.mk

sbrunner: And then I don't know for the others but I will set again the default value....
I the goal it to disable the print setting `PRINT_VERSION = NONE` Is a better way ...
see also: http://docs.camptocamp.net/c2cgeoportal/master/integrator/make.html#makefile-config-variables

jaykayone: OK can we set PRINT_VERSION = NONE in the geoportailv3.mk for now and add to it to main.mk?
So the developers are not bothered with print issues

rmichaelis: I've just tried the print_version but I still have the problem.

sbrunner: Effectively, we have an older version of the CONST_Makfile, commit like it and I will change it when I do an upgrade.

PR #286, opened 2015-02-27
first version of the new print design

jaykayone:


print/print-apps/geoportailv3/config.yaml
```diff
@@ -1,49 +1,349 @@
 pdfConfig: !pdfConfig
   compressed: false
   author: "Mapfish Print"
-  subject: "Mapfish Print"
+  subject: "Mapfish Print Verbose example"
```
sbrunner: Not sure that's the wanted subject ...

print/print-apps/geoportailv3/config.yaml
```diff
@@ -1,49 +1,349 @@
...
 
 throwErrorOnExtraParameters: true
 
 templates:
   A4 portrait: !template
     pdfConfig: !pdfConfig
-      title: "A4 Portrait"
-    reportTemplate: report.jrxml
-    tableData: jrDataSource
+      title: "a4_portrait"
```
sbrunner: Not sure that's the wanted title

jaykayone: they will be localized, this is a translation key

main.mk
```diff
@@ -1,4 +1,5 @@
 INSTANCE_ID = main 
 VARS_FILE = vars_main.yaml
+PRINT_OUTPUT = /var/lib/tomcat/webapps
```
jaykayone: the make file now copies the war to the correct directory, but the created file is 700 and tomcat won't unpack it unless i set it to 755 manually...

print/print-apps/geoportailv3/config.yaml
sbrunner: I should miss something because I don't understand how it can work ...

jaykayone: My understanding (might be wrong though):
The server will send these titles to the client, where they can be translated using ngTranslate.

jaykayone: OK, I think I understood. What will be translated is the name of the template. This one is going into the PDF.
I'll update them

jaykayone:
@sbrunner Ok to be merged? I think the 700 vs 755 issue should be handeled by you in another PR. If you are OK with this, I will create a separate issue

sbrunner:
For the right issue it will be fix in the next version :-)

sbrunner:
Looks good for me

PR #291, opened 2015-03-02
Use "layers" node attribute as WMS "layers" param

### LINKED ISSUE #140: layer_external_wms -> wrong request parameters for "LAYERS" [c2c]
Restant c2c 01/02/2015 0.25j

----

The WMS request that is issued is not correct:

eg for env_sentiers the following request is sent:

http://ws.geoportail.lu/mymaps?category_id=15&SERVICE=WMS&VERSION=1.3.0&REQUEST=GetMap&FORMAT=image%2Fpng&TRANSPARENT=true&LAYERS=env_sentiersnature&CRS=EPSG%3A3857&STYLES=&WIDTH=1680&HEIGHT=892&BBOX=640024.3755957562%2C6361959.241609995%2C704231.4793553044%2C6396050.156225183

LAYERS should be taken from the database and its value is "category" in this case (DB is correct)

elemoine:
When creating a WMS layer the `layers` node attribute should be used as the `layers` parameter in WMS GetMap requests.

Please review.

Fixes #140.

### DIFF
--- a/geoportailv3/static/js/layerfactoryservice.js
+++ b/geoportailv3/static/js/layerfactoryservice.js
@@ -22,7 +22,7 @@ app.GetWmtsLayer;
 
 
 /**
- * @typedef {function(string, string=):ol.layer.Image}
+ * @typedef {function(string, string, string=):ol.layer.Image}
  */
 app.GetWmsLayer;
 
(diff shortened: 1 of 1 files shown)


pgiraud:
Looks good to me. You can merge.

PR #292, opened 2015-03-02
Fix font awesome path

### LINKED ISSUE #289: URLs to fonts are broken again
This is what I see in the debugger: `http://localhost:5000/proj/node_modules/font-awesome/fonts/fontawesome-webfont.woff?v=4.2.0`, which leads to 404s.

elemoine:
This PR fixes the path to font-awesome fonts.

See https://github.com/Geoportail-Luxembourg/geoportailv3/commit/a6b8e9edafbec42245cb8706ba7b41d196ddc228#commitcomment-9981500.

Fixes #289.

### DIFF
--- a/geoportailv3.mk
+++ b/geoportailv3.mk
@@ -19,6 +19,9 @@ CONFIG_VARS += ldap
 
 UTILITY_HELP = "- update-translations		Synchronize the translations with Transifex"
 
+# Add rule that copies the font-awesome fonts to the static/build directory.
+POST_RULES = .build/fonts.timestamp
+
 include CONST_Makefile
 
 DEV_REQUIREMENTS += git+https://github.com/transifex/transifex-client.git@fix-proxies#egg=transifex-client-proxies
@@ -31,3 +34,8 @@ update-translations: $(PACKAGE)/locale/$(PACKAGE)-server.pot $(PACKAGE)/locale/$
 	.build/venv/bin/tx pull -l fr
 	.build/venv/bin/tx pull -l de
 	.build/venv/bin/tx pull -l lb
+
+.build/fonts.timestamp: .build/node_modules.timestamp
+	mkdir -p $(PACKAGE)/static/build/fonts
+	cp node_modules/font-awesome/fonts/* $(PACKAGE)/static/build/fonts/
+	touch $@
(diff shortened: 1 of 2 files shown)


elemoine:
Based on the discussion with @pgiraud in https://github.com/Geoportail-Luxembourg/geoportailv3/commit/a6b8e9edafbec42245cb8706ba7b41d196ddc228 I now understand that the broken path to the font-awesome fonts is related to c2cgeoportal's "cache-busting" again. The cache buster indeed adds an element (the hash) to the path in the static URLs. So we need another `..` to be able to reach the `node_modules` path.

Also, it is important to note that we will break it again if we use the cache buster for the `node_modules` static view. Generated static URLs without using `request.static_url` is not robust. We should not do it. But, at this point, I don't know how we can do it for the fonts path.

pgiraud:
What about copying (or linking) the fonts to a dedicated directory in static folder using a make target?

elemoine:
> What about copying (or linking) the fonts to a dedicated directory in static folder using a make target?

That's an option. But we'd still not use `request.static_url` for creating the fonts URL, which is fragile.

pgiraud:
I we want to use `request.static_url` to build path to fonts in css files,
this means that we need to serve css files as mako templates. This is not
option in my opinion.
URL to resources are usually relative to the css file. If the resources are
close to the css file, it should be easy to manage and not that fragile.

On Mon, Mar 2, 2015 at 11:01 AM, Éric Lemoine <notifications@github.com>
wrote:

> What about copying (or linking) the fonts to a dedicated directory in
> static folder using a make target?
>
> That's an option. But we'd still not use request.static_url for creating
> the fonts URL, which is fragile.
>
> —
> <https://github.com/Geoportail-Luxembourg/geoportailv3/pull/292#issuecomment-76685260>
> .
>



-- 
Pierre GIRAUD
Géomaticien, Analyste

Camptocamp France SAS
Savoie Technolac, BP 352

Tel : 00 33 4 79 44 44 93
Mail : pierre.giraud@camptocamp.com
http://www.camptocamp.com

elemoine:
> If the resources are close to the css file, it should be easy to manage and not that fragile.

I can tell that you were not the one trying to fix https://github.com/Geoportail-Luxembourg/geoportailv3/pull/279 :-) Just kidding…

I agree with you that copying the fonts dir is our best option at this point.

elemoine:
@pgiraud, what do you think about a196a2b?

pgiraud:
This looks very good to me. Please merge.

PR #295, opened 2015-03-03
Disable the measure tools when the measure menu is closed

elemoine:


pgiraud:
Looks good to me.

elemoine:
@rmichaelis do you agree with this change as well?

rmichaelis:
Yes of course I agree. 
Looks good to me.

PR #320, opened 2015-03-04
Disable measure tools when measure menu is closed

### LINKED ISSUE #319: Measure tools do not deactivate when toolbar is closed

elemoine:
This commit fixes a bug where the measure tools were not disabled when closing the measure menu. The problem only exists in non-debug mode. It is related to the "active" property being renamed by the compiler. The fix involves using setActive instead of setting the "active" property directly.

Fixes #319


petzlux:
Looks good to me!

PR #321, opened 2015-03-04
elevationservice is queried only when infobar is open

petzlux:


geoportailv3/static/js/infobar/elevationdirective.js
```diff
@@ -44,25 +44,29 @@ app.module.directive('appElevation', app.elevationDirective);
...
  * @param {angular.$http} $http
+ * @param {angular.Scope} $scope
  * @param {ngeo.Debounce} ngeoDebounce
  * @param {string} elevationServiceUrl
  */
 app.ElevationDirectiveController =
-    function($http, ngeoDebounce, elevationServiceUrl) {
+    function($http, $scope, ngeoDebounce, elevationServiceUrl) {
   var map = this['map'];
+  this['parentCtrl'] = $scope['$parent']['ctrl'];
```
elemoine: Here you make the assumption that the parent controller name is `ctrl`. It's an assumption you should not make in my opinion. There must be a better way.

petzlux:
@elemoine you were right, I think there is a better way !

geoportailv3/static/js/infobar/elevationdirective.js
```diff
@@ -25,7 +25,8 @@ app.elevationDirective = function() {
   return {
     restrict: 'E',
     scope: {
-      'map': '=appElevationMap'
+      'map': '=appElevationMap',
+      'visible': '=appVisible'
```
elemoine: I'd suggest using `appElevationVisible` or `appElevationActive`. The latter is what we use for the measure directive.

petzlux: okay done

elemoine:
Looks much better. I'd just suggest to use `appElevationActive` instead of `appVisible`. Currently, we use the directive name (`appElevation`) as the prefix of directive attributes (e.g. `appElevationActive`).

elemoine:
I would preferred `appElevationActive` for consistency with the measure directive, but it's really up to you. Please merge when the Travis build passes.

petzlux:
okay changed variable name, will merge once travis passed

Issue #211, opened 2015-02-02
QR code dans ngeo [c2c]


sbrunner:
Get from #36 

Restant c2c 01/01/2015 1j
Restant c2c 01/02/2015 1j

jaykayone:
is this going to be similar to https://project.camptocamp.com/viewvc/geoportail_luxembourg/trunk/geoadmin/geoadmin/geoadmin/controllers/shorten.py?view=markup ? 

In my implementation of the print template, I need a controller that generates an image for the QR code. Is it already in c2cgeoportal, will it be, or should I add a controller by myself?

sbrunner:
With ngeo the QR code will be probably generate in client side :-) (=> ngeo only)

jaykayone:
So I will have to write a controller nevertheless to get a qrcode in the print template, is this correct?

sbrunner:
I asked the question to the print developers: https://github.com/mapfish/mapfish-print/issues/199

jaykayone:
implementing a qrcode - controller is really not complicated. I am not sure we should "waste time" to enhance ngeo and mfprint for this.

elemoine:
Décision : l'équipe Luxembourg implémente un controller qrcode dans le projet geoportailv3.

PR #322, opened 2015-03-04
added a controller for generating qrcodes

jaykayone:
controller for generating qrcodes for geoportail urls

### DIFF
--- a/geoportailv3/views/qr.py
+++ b/geoportailv3/views/qr.py
@@ -23,7 +23,7 @@ class Qr(object):
             return HTTPBadRequest()
         if not self.regex.match(url):
             return HTTPNotAcceptable(
-                    "not a valid url for this grcode generator")
+                    "not a valid url for this QR code generator")
         qr = qrcode.QRCode(
                  version=1,
                  error_correction=qrcode.constants.ERROR_CORRECT_L,
--- a/geoportailv3/views/qr.py
+++ b/geoportailv3/views/qr.py
@@ -13,14 +13,17 @@ class Qr(object):
 
     def __init__(self, request):
         self.request = request
-        self.regex=re.compile('^http[s]{0,1}://.*[g-o|geoportal|geoportail]{1}.lu.*$')
+        self.regex = re.compile(
+                '^http[s]{0,1}://.*[g-o|geoportal|geoportail]{1}.lu.*$')
+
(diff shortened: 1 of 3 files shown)


jaykayone:
Closes #211

sbrunner:
Looks good for me :-)

geoportailv3/views/qr.py
```diff
@@ -0,0 +1,39 @@
...
+                '^http[s]{0,1}://.*[g-o|geoportal|geoportail]{1}.lu.*$')
+
+    @view_config(route_name='qr')
+    def getqrcode(self):
+        url = self.request.params.get('url', '')
+        if len(url) == 0:
+            return HTTPBadRequest()
+        if not self.regex.match(url):
+            return HTTPNotAcceptable(
+                    "not a valid url for this grcode generator")
```
pgiraud: grcode -> qrcode

pgiraud: even better "QR code"

elemoine:
@sbrunner where should the package version be set in this case? qrcode is geoportailv3-specific package dependency…

sbrunner:
two possible way :
The easiest is to fix it in the `setup.py`.
Or create a `version.txt` file and add in the `geoportailv3.mk` file:
`REQUIREMENTS += -r version.txt` :-)

elemoine:
+1 for fixing the version in `setup.py`. I don't see the advantage of having a `versions.txt` file here.

jaykayone:
do I understand this correctly?
should I add ==5.1 in the setup.py ?

elemoine:
@jaykayone yes!

jaykayone:
Done, shouldn't this also be dont for elastisearch, python_ldap etc?

jaykayone:
can this be merged?

sbrunner:
For me yes :-)

elemoine:
It looks like this broke the build. See https://travis-ci.org/Geoportail-Luxembourg/geoportailv3/builds/53660413.

jaykayone:
where do you see a correlation to my merge? Travis was OK before I merged
I think the build is broken because of #327 and #328

jaykayone:
OK I see the message now, sorry. 
No idea why it has a problem now but not when I merged ...

PR #323, opened 2015-03-05
added controller to prepare legends for print

### LINKED ISSUE #314: print legends
find a way to include our legends in print.
an idea might be this:
https://project.camptocamp.com/viewvc/geoportail_luxembourg/trunk/geoadmin/geoadmin/geoadmin/controllers/print.py?view=markup

especially this part: 
  weasyprint.HTML(url).write_pdf(legend_buffer, stylesheets=[css])

jaykayone:
closes #314

### DIFF
--- a/geoportailv3/views/preparelegend.py
+++ b/geoportailv3/views/preparelegend.py
@@ -27,9 +27,9 @@ class LegendPreparer(object):
         html = response.read()
         document = html5lib.parse(html)
         for img in document.iter("{http://www.w3.org/1999/xhtml}img"):
-                img.set("src", urljoin(url, img.get("src")))
+            img.set("src", urljoin(url, img.get("src")))
         for link in document.iter("{http://www.w3.org/1999/xhtml}link"):
-                link.set("href", urljoin(url, link.get("href")))
+            link.set("href", urljoin(url, link.get("href")))
         headers = {"Content-Type": 'text/html'}
         output = html5lib.serialize(document,
                                     quote_attr_values=True,
--- a/geoportailv3/views/preparelegend.py
+++ b/geoportailv3/views/preparelegend.py
@@ -32,8 +32,8 @@ class LegendPreparer(object):
                 link.set("href", urljoin(url, link.get("href")))
         headers = {"Content-Type": 'text/html'}
         output = html5lib.serialize(document,
-                quote_attr_values=True,
-                omit_optional_tags=False,
-                use_trailing_solidus=True,
-                quote_char=u"'")
+                                    quote_attr_values=True,
+                                    omit_optional_tags=False,
+                                    use_trailing_solidus=True,
+                                    quote_char=u"'")
         return Response(output, headers=headers)
--- a/geoportailv3/views/preparelegend.py
+++ b/geoportailv3/views/preparelegend.py
@@ -3,9 +3,6 @@ from pyramid.response import Response
 from pyramid.httpexceptions import HTTPBadRequest
 import logging
 import urllib2
-from urlparse import urljoin
-import html5lib
-
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


jaykayone:
the generated html is not valid. Currently looking for a solution

jaykayone:
@sbrunner : have you already used html5lib in prod? Any idea why it drops some quotes around strings and leaves out the closing tags?

I switched to a text replacement approach. Now it produces at least some valid HTML. Trying how to use it in jasper right now. Currently it seems that jasper does not use the stylesheet information.

sbrunner:
You can add `quote_attr_values=True` here: 
https://github.com/Geoportail-Luxembourg/geoportailv3/commit/0b24717fdc1ff33b65f9fd0a1330456ff854c77c#diff-78a846343a2360ffcc5466290f2bb936R34

jaykayone:
Thanks @sbrunner that's what I was looking for.
Now the preparation works as expected. It may have to be adapted later though as I have not managed to include styling in HTML elements in jasper reports. So the preparation may change a bit later on.

jaykayone:
@sbrunner after a lot of tests, my conclusion is that we cannot use the HTML element in jasper reports.
It does not always render images included in the HTML fragment and does not apply any styling.
We have to find another way to include our legends.

Maybe we have to something similar to what has been done in geoportailv2, where weasyprint has been used to convert our html documents to pdf and then combine the different pdfs into one.

Isn't this something that should be addressed in mfprint v3?

Finally, this PR does not have to be merged (yet), as long as we do not have a solution that works for rendering the legends.

jaykayone:
for reference: https://project.camptocamp.com/viewvc/geoportail_luxembourg/trunk/geoadmin/geoadmin/geoadmin/controllers/print.py?revision=1947&content-type=text%2Fplain

geoportailv3/views/preparelegend.py
```diff
@@ -0,0 +1,39 @@
...
+        if len(legendname) == 0:
+            return HTTPBadRequest()
+        url = 'http://wiki.geoportail.lu/doku.php?\
+              id=%s:legend:%s&do=export_html' % (lang, legendname)
+
+        response = urllib2.urlopen(url)
+        html = response.read()
+        document = html5lib.parse(html)
+        for img in document.iter("{http://www.w3.org/1999/xhtml}img"):
+                img.set("src", urljoin(url, img.get("src")))
```
elemoine: Extra indent here, no?

geoportailv3/views/preparelegend.py
```diff
@@ -0,0 +1,39 @@
...
+        for link in document.iter("{http://www.w3.org/1999/xhtml}link"):
+                link.set("href", urljoin(url, link.get("href")))
```
elemoine: Extra indent?

geoportailv3/views/preparelegend.py
```diff
@@ -0,0 +1,39 @@
...
+log = logging.getLogger(__name__)
+
+
+class LegendPreparer(object):
+
+    def __init__(self, request):
+        self.request = request
+
+    @view_config(route_name='preparelegend')
+    def preparelegend(self):
```
elemoine: Note that you don't need a class in this case, you could just use a function:

```py
@view_config(route_name='preparelegend')
def preparelegend(request):
    # …
    return Response(…)
```

Just a comment, feel free to use a class if you prefer.

jaykayone:
@elemoine you are right with your indent remarks.
I don't understand though why travis has accepted them.

in this case it is not very important, as the PR will probably be rejected..

elemoine:
What's the status of this PR? Should we close it?

geoportallux:
yes, but open an issue for the legend printing ...

elemoine:
Ok, I did not follow that part…

geoportallux:
read the comments in the PR :)

elemoine:
> read the comments in the PR :)

Got it!

PR #326, opened 2015-03-09
Removing unused getInfo method

pgiraud:
As far as I understand the `CatalogController::getInfo` is not used.
Let's remove it.
Please review.


elemoine:
It looks like this broke the build. See https://travis-ci.org/Geoportail-Luxembourg/geoportailv3/builds/53660413.

elemoine:
It looks like https://github.com/Geoportail-Luxembourg/geoportailv3/pull/322 broke the build. But your change looks good to me.

elemoine:
Merging this on behalf of @pgiraud.

PR #334, opened 2015-03-10
Fix the c2cgeoportal version

sbrunner:


elemoine:
It looks good to me.

Curious to understand what the problem was though…

sbrunner:
New functionality in c2cgeoportal not compatible with the accrual state of the project :-)
I'm going to upgard it :-)

PR #335, opened 2015-03-10
Lux profile

rmichaelis:
I'm not sure but maybe I should use a service instead of a directive. 

Please review.


elemoine:
Thanks for the PR. I'll review this this afternoon.

elemoine:
Why don't you use `ol.interaction.Draw` (in line string mode) instead of adding your own profile interaction?

geoportailv3/static/js/profile/profiledirective.js
```diff
@@ -0,0 +1,238 @@
...
+  };
+
+  /**
+   * @type {Object}
+   */
+  var extractor = {z: z, dist: dist};
+
+  // Using closures for hoverCallback and outCallback since
+  // wrapping in angular.bind leads to a closure error.
+  // See PR https://github.com/google/closure-compiler/pull/867
```
elemoine: You wouldn't have that problem if you used `goog.bind`.

geoportailv3/templates/index.html
```diff
@@ -114,8 +114,8 @@ <h2 translate>infos</h2>
       <div id="map" ngeo-map="mainCtrl.map">
         <app-infobar app-infobar-map="::mainCtrl.map"></app-infobar>
       </div>
+      <app-profile app-profiledata="mainCtrl.profiledata" app-profile-open="mainCtrl.profileOpen" app-profile-map="mainCtrl.map"></app-profile>
```
elemoine: You want to use one-timing binding fo `app-profile-map`.

geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -120,6 +154,39 @@ app.MeasureController = function($scope, ngeoDecorateInteraction) {
...
+        var req = $.param({
+          'geom': new ol.format.GeoJSON().writeGeometry(geom, encOpt),
+          'nbPoints': 100,
+          'layers': 'dhm'
+        });
+        var config = {
+          headers: {'Content-Type': 'application/x-www-form-urlencoded'}
+        };
+        this['profileOpen'] = true;
+        $http.post('profile.json', req, config).then(
```
elemoine: What's `profile.json` exactly?

rmichaelis: profile.json is the webservice that returns the profile for a selected line.

elemoine: So I guess we should use `appModule.constant` in the `index.html` template like for the other web services. See https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/templates/index.html#L188-L193.

rmichaelis:
> Why don't you use `ol.interaction.Draw` (in line string mode) instead of adding your own profile interaction?

@elemoine
There are essentially two reasons.
The first is that it seemed easier for me (perhaps wrongly) to interact with the event directly from interaction and thus to inherit all existing mechanisms provided by the base. Furthermore at first I wanted to extend `ngeo.interaction.MeasureLength`, because this interaction was doing almost all I wanted. But that was not possible.

The second is that other measurement tools were created in that way. Then I did in the same way, to keep a kind of consistency.

What would be the benefits of not using my own interaction? Performances?

elemoine:
> @elemoine
> There are essentially two reasons.
> The first is that it seemed easier for me (perhaps wrongly) to interact with the event directly from interaction and thus to inherit all existing mechanisms provided by the base. Furthermore at first I wanted to extend ngeo.interaction.MeasureLength, because this interaction was doing almost all I wanted. But that was not possible.
> 
> The second is that other measurement tools were created in that way. Then I did in the same way, to keep a kind of consistency.
> 
> What would be the benefits of not using my own interaction? Performances?

I just thought that using an `ol.interaction.Draw` would be sufficient for the "profile" functionality. On the current http://map.geoportail.lu/ a simple line is drawn: there's no measurement values displayed as the user is drawing the line. Do you want a different behavior that the current one?

elemoine:
> I just thought that using an ol.interaction.Draw would be sufficient for the "profile" functionality. On the current http://map.geoportail.lu/ a simple line is drawn: there's no measurement values displayed as the user is drawing the line. Do you want a different behavior that the current one?

And even if we want to display measurement values as the user is drawing the profile line on the map why not use `ngeo.interaction.MeasureLength` instead of adding a custom interaction to the geoportailv3 code base?

rmichaelis:
> And even if we want to display measurement values as the user is drawing the profile line on the map why not use `ngeo.interaction.MeasureLength` instead of adding a custom interaction to the  geoportailv3 code base?


When I started the development, all my needs were not fulfilled with the version of  `ngeo.interaction.MeasureLength`. Now I'm able to use this interaction. 
But there is still a problem : the feature is not properly propagated with the `ngeo.MeasureEventType.MEASUREEND` event. 
I will open an issue for that.

rmichaelis:
It will be ok as soon as #370 is fixed. 
Please review

geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -31,7 +32,9 @@ app.measureDirective = function(appMeasureTemplateUrl) {
     restrict: 'E',
     scope: {
       'map': '=appMeasureMap',
-      'active': '=appMeasureActive'
+      'active': '=appMeasureActive',
+      'profiledata': '=appProfiledata',
```
elemoine: Rename `appProfiledata` to `appMeasureProfiledata` and `appProfileOpen` to `appMeasureProfileopen` for consistency.

geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -120,6 +162,35 @@ app.MeasureController = function($scope, ngeoDecorateInteraction) {
   ngeoDecorateInteraction(measureAzimut);
   map.addInteraction(measureAzimut);
 
+  measureProfile.on(ngeo.MeasureEventType.MEASUREEND, goog.bind(function(evt) {
```
elemoine: We use `goog.events.listen` everywhere else. In this case:

```js
goog.events.listen(measureProfile, ngeo.MeasureEventType.MEASUREEND,
    /**
     * @param {ngeo.MeasureEvent} evt Measure event.
     */
    function(evt) {
      // …
    }, undefined, this);
```

geoportailv3/static/js/profile/profiledirective.js
```diff
@@ -0,0 +1,237 @@
+/**
+ * @fileoverview This file provides a measure directive. This directive is used
+ * to create a measure panel in the page.
+ *
+ * Example:
+ *
+ * <app-measure app-measure-map="::mainCtrl.map"></app-measure>
```
elemoine: The example refers to the `app-measure` directive, which is not the directive this file includes.

geoportailv3/static/js/profile/profiledirective.js
```diff
@@ -0,0 +1,237 @@
...
+/**
+ * @param {string} appProfileTemplateUrl Url to layermanager template
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+app.profileDirective = function(appProfileTemplateUrl) {
+  return {
+    restrict: 'E',
+    scope: {
+      'profiledata': '=appProfiledata',
```
elemoine: `appProfiledata` should be changed to `appProfileData`.

geoportailv3/static/js/profile/profiledirective.js
```diff
@@ -0,0 +1,237 @@
...
+app.ProfileController = function($scope) {
+  /**
+   * @private
+   */
+  this.distanceLabel_ = 'Distance : ';
+
+  /**
+   * @private
+   */
+  this.elevationLabel_ = 'Elevation : ';
```
elemoine: Empty line missing after this line.

elemoine:
I've just added a few minor comments, otherwise looks good to me.

pgiraud:
@rmichaelis Can you please rebase this branch and fix the indentation issues?
For more chances to have travis happy you should run the check command.
If you don't have time for this, I can try to fix the indentation issues. Please tell me.

rmichaelis:
May I merge this PR ?

geoportailv3/static/js/profile/profiledirective.js
pgiraud: @rmichaelis have you tried using goog.bind as Éric suggested?

pgiraud:
@rmichaelis did you see Éric' comment about using `goog.bind`?

jaykayone:
@rmichaelis is on vacation until tomorrow

pgiraud:
I propose we merge this for him and address remaining points later.

jaykayone:
Ok

PR #336, opened 2015-03-10
Selected Layers reorder using drag & drop

petzlux:
~~Only working in debug mode, help appreciated~~ see comment

### DIFF
--- a/geoportailv3/static/js/layermanager/layermanagerdirective.js
+++ b/geoportailv3/static/js/layermanager/layermanagerdirective.js
@@ -14,8 +14,8 @@
  * The content of the array of layers may change, but not the array itself.
  */
 goog.provide('app.layermanagerDirective');
-
 goog.require('app');
+goog.require('ngeo.sortableDirective');
 
 
 /**
@@ -57,6 +57,7 @@ app.LayermanagerController = function() {
  */
 app.LayermanagerController.prototype.removeLayer = function(layer) {
   this['map'].removeLayer(layer);
+
 };
 
 
--- a/geoportailv3/static/js/maincontroller.js
+++ b/geoportailv3/static/js/maincontroller.js
@@ -11,6 +11,7 @@ goog.provide('app.MainController');
 
 goog.require('app');
 goog.require('app.LocationControl');
+goog.require('ngeo.SyncArrays');
 goog.require('ngeo.mapDirective');
 goog.require('ol.Map');
 goog.require('ol.View');
@@ -29,12 +30,13 @@ goog.require('ol.tilegrid.WMTS');
  * @param {angularGettext.Catalog} gettextCatalog Gettext catalog.
  * @param {Object.<string, string>} langUrls URLs to translation files.
  * @param {Array.<number>} defaultExtent Default geographical extent.
+ * @param {ngeo.SyncArrays} ngeoSyncArrays
  * @constructor
  * @export
  * @ngInject
  */
 app.MainController = function($scope, gettextCatalog, langUrls,
-    defaultExtent) {
+    defaultExtent, ngeoSyncArrays) {
 
   /**
    * @type {angularGettext.Catalog}
@@ -106,7 +108,7 @@ app.MainController = function($scope, gettextCatalog, langUrls,
 
   this.setMap_();
   this.switchLanguage('fr');
-  this.manageSelectedLayers_($scope);
+  this.manageSelectedLayers_($scope, ngeoSyncArrays);
 };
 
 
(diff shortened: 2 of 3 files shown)


geoportailv3/static/js/layermanager/layermanagerdirective.js
```diff
@@ -44,10 +45,24 @@ app.module.directive('appLayermanager', app.layermanagerDirective);
...
+ * @param {ngeo.SyncArrays} ngeoSyncArrays
  * @export
  */
-app.LayermanagerController = function() {
+app.LayermanagerController = function($scope, ngeoSyncArrays) {
   this['uid'] = goog.getUid(this);
+  ngeoSyncArrays(this['map'].getLayers().getArray(),
+      this['layers'], true, $scope,
+      goog.bind(function(layer) {
+        return goog.isDefAndNotNull(layer['values_']['metadata']);
```
petzlux: Is there a better way to check if a layer is a baselayer or not ? couldnt find a specific property

pgiraud: Testing that the layer is first in the list should be enough (ie. index == 0). This is what we do elsewhere.

elemoine: The use of `ngeoSyncArrays` to synchronize the array of layers in the map with the array of selected layers should be in the main controller (in place of what currently is in the `manageSelectedLayers_` method) . Indeed, the layer manager should not know about background layers, so it should not know how to filter out background layers.

geoportailv3/static/js/layermanager/layermanagerdirective.js
```diff
@@ -44,10 +45,24 @@ app.module.directive('appLayermanager', app.layermanagerDirective);
...
+      }, this)
+  );
+  $scope.$watchCollection(function() {
```
elemoine: Aren't you missing a `goog.bind` here?

petzlux:
I had to do a cleanall before the changes would work in production mode. @rmichaelis told me he had same problem beforehand. Is this because partials  are not rebuild after a change ? @elemoine

elemoine:
> Is this because partials are not rebuild after a change ? @elemoine

After a change to what?

geoportailv3/static/js/maincontroller.js
```diff
@@ -151,6 +154,19 @@ app.MainController.prototype.manageSelectedLayers_ = function(scope) {
...
+        return goog.array.indexOf(
+            this['map'].getLayers().getArray(), layer) !== 0;
+      }, this)
+  );
+  scope.$watchCollection(goog.bind(function() {
+    return this['selectedLayers'];
+  }, this), goog.bind(function() {
+    this['map'].render();
+  }, this));
+
```
elemoine: You need to remove the old code in the `manageSelectedLayers_`. Your code should replace that old code.

petzlux:
@elemoine thanks for comment, I removed unnecessary code.

petzlux:
After having done changes on maincontroller  , in debug mode the changes would be applied, not in production mode (even after redoing make serve )

elemoine:
> After having done changes on maincontroller , in debug mode the changes would be applied, not in production mode (even after redoing make serve )

This is strange, I've never experienced that. I'll look at it this afternoon, to see if I can reproduce the problem.

elemoine:
Is this patch works both in prod and debug modes then I think it's good to go!

petzlux:
@rmichaelis had the same problem where he had to do a cleanall for changes to apply. Maybe he can say exactly where. 

before merging still need to apply css to element while dragged.

rmichaelis:
I've created an html used by a directive. 
After testing it in both debug and production mode, I decided to modify it. 
The changes were available in debug mode but not in production mode.
A simple clean didn't change the behavior. 
I had to execute a cleanall to have my html available in production.

elemoine:
> 've created an html used by a directive.
> After testing it in both debug and production mode, I decided to modify it.
> The changes were available in debug mode but not in production mode.
> A simple clean didn't change the behavior.
> I had to execute a cleanall to have my html available in production.

`make build` should be sufficient after modifying a directive partial. Can you please confirm that it was not sufficient in your case?

rmichaelis:
Yes I confirm.

elemoine:
> Yes I confirm.

Thanks. I'll fix it.

petzlux:
@pgiraud At the moment, the dragged selected Layer does not have all CSS rules applied. I would want it to look the same as a normal selected Layer. 
I looked at the .less CSS selector structure, and it relies for the panel rules to be inside a ul .layermanager
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/less/theme.less#L112
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/less/layermanager.less#L4

But the draggable element is added to the DOM at the very end outside the ul.layermanager. How can we apply the same styling without repeating rules ? 

image

PR #341, opened 2015-03-11
Set Zoom & Extent Limits

### LINKED ISSUE #306: Set zoom and extent limits
At the moment no zoom or extent limits are set. This causes issues with scaleselector being out of bounds, and user loosing Luxembourg in sea of grey

petzlux:
closes #306


jaykayone:
we should already set the extent limits in order to cover the whole topo map of greater region

pgiraud:
@petzlux can the branch be deleted?

PR #347, opened 2015-03-16
Better style for dragged element in layermanager

### LINKED ISSUE #343: Dragged Element CSS Styling
@pgiraud At the moment, the dragged selected Layer does not have all CSS rules applied. I would want it to look the same as a normal selected Layer. 
I looked at the .less CSS selector structure, and it relies for the panel rules to be inside a ul .layermanager
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/less/theme.less#L112
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/less/layermanager.less#L4

But the draggable element is added to the DOM at the very end outside the ul.layermanager. How can we apply the same styling without repeating rules ? 

image

pgiraud:
Commit message says it all.
screenshot from 2015-03-16 17 00 32

Depends on: https://github.com/camptocamp/ngeo/pull/195

Fixes #343


elemoine:
Please merge.

PR #350, opened 2015-03-18
Groundwork for permalinks

elemoine:
This PR shows how we can read and create permalinks.

The PR adds an app-specific map directive. At init time, this directive reads the zoom level and the view center from the URL. And it updates the URL when the map view changes.

The URL looks like this: `?zoom=11&x=675999&y=6408641`. Note that `x` and `y` are web mercator coordinates, making the URL not compatible with the previous application!

If we decide to store map/app states in the localStorage, we could create an `appStateManager` service that will manage both the URL and the local storage. So instead of using the `ngeoLocation` service directly the application components/directives would use the `appStateManager` service.


elemoine:
`X` and `Y` in the permalink now use the LUREF projection. And `zoom` in the permalink now uses the « old » zoom ranges. There are still bad cases for the zoom levels. I've added FIXMEs for the bad cases in the code.

elemoine:
I've now added an `appStateManager` service that writes the application state to both the URL and the localStorage. The map component now uses that service to read and write the `X`, `Y`, and `zoom` values. For read, values are first read from the URL, and then from the localStorage if there is no value in the URL.

geoportallux:
+1

petzlux:
Could we not have a legacy parameter with default true so that we could mix old (LUREF xyz) and new (Webmercator xyz) style , with a view to sunset oldstyle permalinks at some point in the future ?

jaykayone:
Great idea

elemoine:
> Could we not have a legacy parameter with default true so that we could mix old (LUREF xyz) and new (Webmercator xyz) style , with a view to sunset oldstyle permalinks at some point in the future ?

I like the idea too. Do you agree with the following statements?

* No `legacy` parameter in the URL will be the same as `legacy=true`.
* A non-legacy URL has a `legacy` parameter set to `false`.
* When we receive a legacy URL we will directly convert it into a non-legacy URL (in the address bar).
* We never provide legacy URLs to the user. This means `legacy` is always set to `false` in the URLs/permalinks we generate.

Instead of a `legacy` parameter we could use a `version` parameter, where the `version` defaults to `2` when there's no `version` parameter in the URL. And a non-legacy URL will have `version` set to `3`.

What do you think?

geoportallux:
we agree

elemoine:
Cool. I'll go ahead and implement this.

petzlux:
:+1:

jaykayone:
let's go for version = 2/3

elemoine:
I think I've changed the code as discussed.

elemoine:
And note that there's still a fixme for the v2 to v3 zoom levels mapping.

PR #351, opened 2015-03-18
Add the typeahead.js externs file

elemoine:
With https://github.com/camptocamp/ngeo/pull/187 the ngeox.js externs file depends on externs defined in the typeahead.js externs file. So this requires adding the typeahead.js extens file to the project't build.json configuration file.


elemoine:
The build passes again with this change. @rmichaelis are you ok with me merging this?

rmichaelis:
Yes, I've just tested it and it works fine.

PR #353, opened 2015-03-19
oldOpacity can be managed internally, don't create layer property

pgiraud:
The goal here is to manage the old opacities internally instead of creating a new property in the layer.


geoportailv3/static/js/layermanager/layermanagerdirective.js
```diff
@@ -48,6 +48,13 @@ app.module.directive('appLayermanager', app.layermanagerDirective);
  */
 app.LayermanagerController = function() {
   this['uid'] = goog.getUid(this);
+
+  /**
+   * Hash array to keep track of opacities set on layers.
+   * @type {Object}
```
elemoine: You can be stricter with the type: `Object.<number, number>`.

pgiraud:
Thanks @elemoine. I took your comment into account.

PR #361, opened 2015-03-20
Update pyramid closure

pgiraud:
This will need to be done in the update scaffold of c2cgeoportal as well.


elemoine:
For everyone's information, this fixes a `closure-deps` related problem that has been happening since the update to the latest c2cgeoportal version. The problem is related to the c2cgeoportal `update` scaffold using an old commit of `pyramid_closure` in `CONST_requirements.txt`.

elemoine:
Please merge.

PR #368, opened 2015-03-23
Fix ordering in scale selector

### LINKED ISSUE #342: Ngeo Scale Selector - levels not ordered properly
In the scale selector, the zoom levels and corresponding scales are not ordered properly, i.e 1:1.000.000 comes at bottom when it should be at top of list. 

see  https://github.com/Geoportail-Luxembourg/geoportailv3/blob/petzlux_extents/geoportailv3/static/js/infobar/scaleselectordirective.js#L54-L55

screen shot 2015-03-11 at 16 55 54

elemoine:
Fixes #342.


pgiraud:
OK, looks good.

elemoine:
Note that this requires running `make update-node-modules`.

PR #373, opened 2015-03-24
Use module.service for appThemes

elemoine:
As [noticed](https://github.com/Geoportail-Luxembourg/geoportailv3/commit/9614a7b3fff29b23c3ff00fc000a278afbae3886#commitcomment-10363813) by @pgiraud we can use `module.service` rather than `module.factory` for the `appThemes` service.


pgiraud:
I like simplification. Looks good, please merge.

PR #377, opened 2015-03-24
Fix localStorage/permalink bug

### LINKED ISSUE #372: localstorage
the seems to be a problem with the localstorage:

http://devv3rm.geoportail.lu/?version=3&zoom=12&X=674880&Y=6390772
If you reopen the page without url params:
http://devv3rm.geoportail.lu/?version=3&zoom=8&X=16676259&Y=623889

I think that it does not use the version=3 parameter and then applies the coordinate system switch to the coordinates that are already in webmercator or something like that

elemoine:
Fix a bug where the map directive converts x, y and zoom from version 2 to version 3 when the values are read from the local storage. There should be no conversion when the values are read from the local storage. The conversions occur only when reading from the location URL, when there's no version parameter or when the version parameter is set to 2.

Depends on https://github.com/camptocamp/ngeo/pull/203.

Fixes #372.

### DIFF
--- a/geoportailv3/static/js/map/mapdirective.js
+++ b/geoportailv3/static/js/map/mapdirective.js
@@ -55,7 +55,7 @@ app.MapController = function(appStateManager, ngeoDebounce) {
   /** @type {number} */
   var version = appStateManager.getVersion();
 
-  var zoom = appStateManager.getParam('zoom');
+  var zoom = appStateManager.getInitialValue('zoom');
 
   /** @type {number} */
   var viewZoom;
@@ -66,8 +66,8 @@ app.MapController = function(appStateManager, ngeoDebounce) {
     viewZoom = 8;
   }
 
-  var x = appStateManager.getParam('X');
-  var y = appStateManager.getParam('Y');
+  var x = appStateManager.getInitialValue('X');
+  var y = appStateManager.getInitialValue('Y');
 
   /** @type {ol.Coordinate} */
   var viewCenter;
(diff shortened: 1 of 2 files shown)


elemoine:
https://github.com/camptocamp/ngeo/pull/203 is now merged.

elemoine:
Travis build now passes. Please review.

geoportailv3/static/js/statemanagerservice.js
```diff
@@ -21,28 +21,59 @@ goog.require('ngeo.Location');
...
 
   /**
-   * @type {ngeo.Location}
+   * @type {Object.<string ,string>}
    * @private
    */
-  this.ngeoLocation_ = ngeoLocation;
-
-  var version = ngeoLocation.getParam('version');
+  this.states_ = {};
```
pgiraud: I would name it `initialState_`.

pgiraud:
As we just discussed, please use singular for state.
{{{
updateStates(states) -> updateState(object)
getInitialState(key) -> getInitialValue(key)
}}}

elemoine:
Patch updated as discussed.

elemoine:
Thanks for the review @pgiraud. Merging.

geoportailv3/static/js/map/mapdirective.js
```diff
@@ -81,7 +81,7 @@ app.MapController = function(appStateManager, ngeoDebounce) {
   view.setCenter(viewCenter);
   view.setZoom(viewZoom);
 
-  app.MapController.updateParams_(appStateManager, view);
+  app.MapController.updateStates_(appStateManager, view);
```
pgiraud: Let's be consistent and use `updateState_` here. Sorry I'm picky.

elemoine: You're not picky at all. It's an oversight. Fixing now…

PR #380, opened 2015-03-25
Layer exclusions updates (translation and layer removal)

pgiraud:
Please review.

@jaykayone I'm not sure the messages are clear enough. The messages could probably be shorter and more explicit.
Also though this exclusion notification mechanism improves usability, I'm still not sure users understand immediately what happens. At least, they are notified.

### DIFF
--- a/geoportailv3/static/js/exclusionmanagerservice.js
+++ b/geoportailv3/static/js/exclusionmanagerservice.js
@@ -61,8 +61,8 @@ app.ExclusionManager = function(gettextCatalog, ngeoBackgroundLayerMgr,
    * @type {string}
    * @private
    */
-  this.exclusionMessage2_ = 'The layer(s) <b>{{layersToHide}}</b> have been ' +
-      'deactivated because they cannot be displayed while the layer ' +
+  this.exclusionMessage2_ = 'The layer(s) <b>{{layersToRemove}}</b> have ' +
+      ' been removed because they cannot be displayed while the layer ' +
       '<b>{{layer}}</b> is displayed';
 };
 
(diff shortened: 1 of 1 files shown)


jaykayone:
The messages are OK for me. They could stay a bit longer.
But there is a functional problem:
- Activate cadastral parcels -> background disappears (with notification) (OK)
- reactivate background -> parcels disappear (with notification) (OK)

BUT: 
- checkbox of cadastral parcels stays
AND:
- no possibility to re-display cadastral parcels from that moment on

pgiraud:
Actually the "cadastral parcels" layer is still in the selected layers but it's opacity is set to 0. Is it not the expected behavior?

jaykayone:
OK I see. In that case, the icon with the eye should be visible in the catalog and in the non-collapsed part of the layer in the layer manager, otherwise people will not understand what's going on.
If we don't want to change the interface, I'd suggest to remove the layer rather than deactivating it

pgiraud:
OK.

We may also want to inform that the layer opacity is set to 0 without expanding the layer actions by for example modify the visual aspect.

jaykayone:
My short experience shows me that people tend to stay in the catalog view rather than switching to my_layers (unless they want to reorder or change transparency). so they should also see inside the catalog when a layer is deactivated (maybe a grey checkbox and an additional click brings it back). Sort of a 3-state-checkbox?

pgiraud:
OK. I updated the code so that the layers are removed when not compatible with newly chosen background. This is indeed a more natural behavior.

jaykayone:
closes #379

PR #382, opened 2015-03-25
Show elevation offset in azimuth control

rmichaelis:
Remarks from one of our tester regarding the whole azimuth control :  
title for distance and azimuth would be helpful 
a line from center to north would be helpful too.

Please review.

### DIFF
--- a/geoportailv3/static/js/measure/measuredirective.js
+++ b/geoportailv3/static/js/measure/measuredirective.js
@@ -47,13 +47,27 @@ app.module.directive('appMeasure', app.measureDirective);
 
 /**
  * @param {angular.Scope} $scope Scope.
+ * @param {angular.$q} $q
+ * @param {angular.$http} $http
  * @param {ngeo.DecorateInteraction} ngeoDecorateInteraction Decorate
  *     interaction service.
+ * @param {string} elevationServiceUrl the url of the service
  * @constructor
  * @export
  * @ngInject
  */
-app.MeasureController = function($scope, ngeoDecorateInteraction) {
+app.MeasureController = function($scope, $q, $http, ngeoDecorateInteraction,
+    elevationServiceUrl) {
+
+  /**
+   * @private
+   */
+  this.http_ = $http;
+
+  /**
+   * @private
+   */
+  this.elevationServiceUrl_ = elevationServiceUrl;
 
   var style = new ol.style.Style({
     fill: new ol.style.Fill({
@@ -77,8 +91,9 @@ app.MeasureController = function($scope, ngeoDecorateInteraction) {
 
   /**
    * @type {ol.Map}
+   * @private
    */
-  var map = this['map'];
+  this.map_ = this['map'];
 
   var measureLength = new ngeo.interaction.MeasureLength({
     sketchStyle: style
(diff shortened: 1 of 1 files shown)


geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -47,13 +47,27 @@ app.module.directive('appMeasure', app.measureDirective);
...
  * @ngInject
  */
-app.MeasureController = function($scope, ngeoDecorateInteraction) {
+app.MeasureController = function($scope, $q, $http, ngeoDecorateInteraction,
+    elevationServiceUrl) {
+
+  /**
+   * @private
+   */
+  this.http_ = $http;
```
elemoine: Very minor, but we could use `this.$http_` to make it clear and explicit that it's an Angular service.

geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -47,13 +47,27 @@ app.module.directive('appMeasure', app.measureDirective);
...
+
+  /**
+   * @private
```
elemoine: You usually provide a `@type` to instance properties.

geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -47,13 +47,27 @@ app.module.directive('appMeasure', app.measureDirective);
...
  * @constructor
  * @export
  * @ngInject
  */
-app.MeasureController = function($scope, ngeoDecorateInteraction) {
+app.MeasureController = function($scope, $q, $http, ngeoDecorateInteraction,
+    elevationServiceUrl) {
+
+  /**
+   * @private
```
elemoine: You can add a `@type` here.

geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -118,7 +133,29 @@ app.MeasureController = function($scope, ngeoDecorateInteraction) {
...
+        var geometryCollection =
+            /** @type {ol.geom.GeometryCollection} */
+            (evt.feature.getGeometry());
+
+        var radius =
+            /** @type {ol.geom.LineString} */
+            (geometryCollection.getGeometries()[0]);
+
+        $q.all([this.getElevation_(radius.getCoordinates()[0]),
+              this.getElevation_(radius.getCoordinates()[1])]
```
elemoine: You can avoid calling `radius.getCoordinates()` twice.

geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -134,4 +171,21 @@ app.MeasureController = function($scope, ngeoDecorateInteraction) {
 };
 
 
+/**
+ * @param {ol.Coordinate} coordinates
+ * @return {angular.$q.Promise}
+ * @private
+ */
+app.MeasureController.prototype.getElevation_ = function(coordinates) {
+  var lonlat =
```
elemoine: Any reason for naming this `lonlat`?

elemoine: I mean: is it really a longitude and a latitude that you store in that variable?

elemoine:
I've just added a few minor comments, otherwise looks good to me.

geoportailv3/static/js/measure/measuredirective.js
rmichaelis: @elemoine Yes indeed this is not a longitude and latitude. But I chose the name lonlat, because the elevation webservice is using lon and lat as parameter name.
Thus I've used the same convention, in order to show that the coordinates are in the same reference system than the ones used by the webservice.
In your opinion, what would be a better name ? eastnorth ?

elemoine: > eastnorth ?

Oh yes!

PR #384, opened 2015-03-25
Better looking and more accessible checkboxes

pgiraud:
I worked on the checkboxes in the catalog tree.
The idea was to push the checkboxes to the right. However I didn't find any way to make it correctly accessible.
I strongly think that it's a good thing to link the label (layer name) with the checkbox and group them into one single clickable element. Aligning the checkbox to the right makes things technically harder.
So, though it's a bit less comfortable visually, I think it works very well with what i did.

screenshot from 2015-03-25 17 18 00

### DIFF
--- a/geoportailv3/static/js/catalog/catalogdirective.js
+++ b/geoportailv3/static/js/catalog/catalogdirective.js
@@ -90,4 +90,20 @@ app.CatalogController.prototype.getLayer = function(node) {
 };
 
 
+/**
+ * Add or remove layer from map.
+ * @param {Object} node Tree node.
+ * @export
+ */
+app.CatalogController.prototype.toggle = function(node) {
+  var layer = this.getLayerFunc_(node);
+  var map = this['map'];
+  if (map.getLayers().getArray().indexOf(layer) >= 0) {
+    map.removeLayer(layer);
+  } else {
+    map.addLayer(layer);
+  }
+};
+
+
 app.module.controller('AppCatalogController', app.CatalogController);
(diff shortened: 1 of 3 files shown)


jaykayone:
looks good. label should be clickable too.

pgiraud:
They are! The checkbox and the label are in the same "<a>" element.

jaykayone:
don't know why but neither the checkboxes nor the labels are clickable: 
http://devv3.geoportail.lu/main?version=3&zoom=9&X=669751&Y=6416190

I did a clean of the project before building

pgiraud:
OK. This works in debug mode, not when built.
I need to `@export` the `toggle` method.

jaykayone:
ok in debug mode it works fine

geoportailv3/static/js/catalog/catalogdirective.js
```diff
@@ -90,4 +90,19 @@ app.CatalogController.prototype.getLayer = function(node) {
...
+ * Add or remove layer from map.
+ * @param {Object} node Tree node.
+ * @export
+ */
+app.CatalogController.prototype.toggle = function(node) {
+  var layer = this.getLayerFunc_(node);
+  if (this['map'].getLayers().getArray().indexOf(layer) >= 0) {
+    this['map'].removeLayer(layer);
+  } else {
+    this['map'].addLayer(layer);
```
elemoine: You could use a local `map` variable here.

pgiraud:
OK tests pass and comments are taken into account.

PR #394, opened 2015-04-03
Theme colors

pgiraud:
With this pull request the colors are configured more manually but the result is better.


elemoine:
Look good to me.

PR #395, opened 2015-04-07
Fix compile errors

elemoine:
This PR fixes new compile errors that we now get because we use a more recent version of Closure Compiler.

See also https://github.com/camptocamp/ngeo/pull/207 and https://github.com/openlayers/ol3/pull/3511.

### DIFF
--- a/build.json
+++ b/build.json
@@ -63,7 +63,6 @@
       "missingGetCssName",
       "missingProperties",
       "missingProvide",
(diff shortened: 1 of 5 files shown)


elemoine:
Travis is green, so this does address the compilation errors.

Note that you will need to run `make update-node-modules`, to update `ngeo` and `openlayers` in `node_modules`.

PR #399, opened 2015-04-10
Upgrade to GeoMapFish master

sbrunner:
La localisation côté serveur doit aussi être corrigée ...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 /production.ini
 /project.yaml
 /*.egg-info/
+/print/WEB-INF/classes/logback.xml
 /geoportailv3/locale/*/LC_MESSAGES/*.mo
 /geoportailv3/static/mobile/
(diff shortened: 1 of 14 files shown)


elemoine:
It looks good to me. Please merge when Travis reports success (which is not the case right now).

sbrunner:
I also update the print :-)

elemoine:
It would be good to know what we (geoportailv3 developers) need to do after a c2cgeoportal upgrade? @sbrunner, would you mind adding a few notes here about that?

sbrunner:
Sauf erreur:
```
make -f <user>.mk update
make -f <user>.mk build
```
Devrai suffire
http://docs.camptocamp.net/c2cgeoportal/master/integrator/upgrade_application.html#easy-updating-an-application-code

elemoine:
Thank you!

PR #402, opened 2015-04-13
Use closure-util's build command

### LINKED ISSUE #51: Rely on openlayers's or ngeo's tasks scripts instead of having our own copies [c2c]
Restant c2c 01/02/2015 0.25j
Restant c2c 01/03/2015 0.25j
Restant c2c 01/04/2015 0.25j

elemoine:
With this PR we rely on closure-util 1.4.0's `build` command instead of using specific build scripts. This allows getting rid of many project-specific scripts, which were copied from the ol3 repository. It also makes the js build process faster.

Fixes #51.


elemoine:
@sbrunner, this is one to port to c2cgeoportal, which should be pretty straightforward.

sbrunner:
The `compile-catalog.js` and the `extract-messages.js` shouldn't be moved in ngeo ?
It isn't it a better way to shared them ?

elemoine:
> The compile-catalog.js and the extract-messages.js files shouldn't be moved in ngeo ?

I think they should be moved to the create scaffold.

sbrunner:
Actually they are on the update scaffold..., Can you explain why they shouldn't be in ngeo ?

fredj:
@elemoine `fs-extra`, `jsdoc`, `temp` and `walk` npm packages can removed from package.json

elemoine:
> @elemoine fs-extra, jsdoc, temp and walk npm packages can removed from package.json

You're right. Thanks!

elemoine:
> Actually they are on the update scaffold..., Can you explain why they shouldn't be in ngeo ?

Because at this point, `angular-gettext` is not used by `ngeo`. And, with Angular 1.4 including internationalization support, it's not clear that we will still use `angular-gettext` in the near future.

PR #403, opened 2015-04-13
Reload themes after authentication

### LINKED ISSUE #389: Rechargement du catalogue après login
Vu qu'un utilisateur peut avoir le droit de voir certaines couches uniquement quand il est logué, il faut recharger le catalogue après login

elemoine:
The objective of this PR is to reload themes after the user has logged in or logged out. This is still WIP for now.

Fixes #389.

### DIFF
--- a/geoportailv3/static/js/authentication/authenticationdirective.js
+++ b/geoportailv3/static/js/authentication/authenticationdirective.js
@@ -11,6 +11,9 @@ goog.require('app');
 app.authenticationDirective = function(appAuthenticationTemplateUrl) {
   return {
     restrict: 'E',
+    scope: {
+      'roleId': '=appAuthenticationRoleid'
+    },
     controller: 'AppAuthenticationController',
     controllerAs: 'ctrl',
     bindToController: true,
@@ -66,32 +69,33 @@ app.AuthenticationController = function($http, loginUrl, logoutUrl,
    */
   this.http_ = $http;
 
-  /**
-   * @type {string}
-   */
-  this['login'] = null;
-
 
   /**
-   * @type {string}
+   * @type {string|undefined}
    */
-  this['email'] = null;
+  this['login'] = undefined;
 
 
   /**
-   * @type {string}
+   * @type {string|undefined}
    */
-  this['role'] = null;
+  this['email'] = undefined;
+
 
   /**
-   * @type {number}
+   * @type {string|undefined}
    */
-  this['role_id'] = 0;
+  this['role'] = undefined;
 
   /**
-   * @type {string}
+   * @type {number|undefined}
    */
-  this['name'] = null;
+  this['roleId'] = undefined;
+
+  /**
+   * @type {string|undefined}
+   */
+  this['name'] = undefined;
 
   /**
    * @type {boolean}
@@ -121,12 +125,13 @@ app.AuthenticationController.prototype.authenticate = function(credentials) {
         if (status == 200) {
           that.getUserInfo();
           that['isError'] = false;
-        }else {
+        } else {
           that['isError'] = true;
         }
-      }).error(function(data, status, headers, config) {
-    that['isError'] = true;
-  });
+      }).error(
+      function(data, status, headers, config) {
+        that['isError'] = true;
+      });
 };
 
 
@@ -144,14 +149,15 @@ app.AuthenticationController.prototype.logout = function() {
         if (status == 200) {
           that.getUserInfo();
           that['isError'] = false;
-        }else {
+        } else {
           that.getUserInfo();
           that['isError'] = true;
         }
-      }).error(function(data, status, headers, config) {
-    that.getUserInfo();
-    that['isError'] = true;
-  });
+      }).error(
+      function(data, status, headers, config) {
+        that.getUserInfo();
+        that['isError'] = true;
+      });
 };
 
 
@@ -175,13 +181,13 @@ app.AuthenticationController.prototype.getUserInfo = function() {
               data['mail'],
               data['sn']
           );
-        }else {
-          that.setUserInfo(null, null, null, null, null);
+        } else {
+          that.clearUserInfo();
         }
-
-      }).error(function(data, status, headers, config) {
-    that.setUserInfo(null, null, null, null, null);
-  });
+      }).error(
+      function(data, status, headers, config) {
+        that.clearUserInfo();
+      });
 };
 
 
@@ -190,7 +196,7 @@ app.AuthenticationController.prototype.getUserInfo = function() {
  * @export
  */
 app.AuthenticationController.prototype.isAuthenticated = function() {
-  if (goog.isDefAndNotNull(this['login']) && this['login'].length > 0) {
+  if (goog.isDef(this['login']) && this['login'].length > 0) {
     return true;
   }
   return false;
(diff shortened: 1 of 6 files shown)


elemoine:
This is now ready for review.

With no LDAP access I haven't been able to fully test that PR. I think it would good if someone from the Luxembourg could test my changes before we merge that PR.

When an anonymous user loads the application in the browser there should one request to the `themes` service and only one.

When an already authenticated user loads the application in the browser there should be two requests to the `themes` service. The first one is done eagerly, because, as discussed with @jaykayone in #389, in most cases, users are not yet logged in when they load the application.

When the user logs in there should be one request to the `themes` service, and the `themes` service should see an "authenticated" user.

When the user logs out there should be one request to the `themes` service, and the `themes` service should see an "anonymous" user.

Please test and review.

rmichaelis:
I can't test.
The build fails due to the proxy error due to the git protocol.
I still don't know how to bypass this problem.

elemoine:
Could you please test with patch:

```diff
diff --git a/CONST_requirements.txt b/CONST_requirements.txt
index 329b2c4..f229fc2 100644
--- a/CONST_requirements.txt
+++ b/CONST_requirements.txt
@@ -3,5 +3,5 @@
 --find-links http://pypi.camptocamp.net/internal-pypi/index/c2cgeoportal-win
 -r CONST_versions.txt
 -e git+https://github.com/camptocamp/pyramid_closure@819bc43420b3cd924d8698c5a9606592c19dbb15#egg=pyramid_closure
--e git+https://github.com/Pylons/pyramid@1e02bbfc0df09259bf207112acf019c8dba44a90#egg=pyramid
+https://github.com/Pylons/pyramid/archive/1e02bbfc0df09259bf207112acf019c8dba44a90.zip#egg=pyramid
 -e .
```

rmichaelis:
Thanks for the patch. It woks fine. 
I'm now able to test the PR.

elemoine:
> Thanks for the patch. It woks fine.
> I'm now able to test the PR.

So I'll create a PR with this. Thanks for testing.

elemoine:
@rmichaelis I now set `cache` to `false` when reloading themes with `$http.get`. I hope this will fix the issue. Can you please test again?

elemoine:
As discussed with @sbrunner we should use a `role` parameter in the URL.

elemoine:
Patch updated.

So, when the role changes the themes are reloaded, and the role id is sent to the `themes` service using `&role=<role_id>` in the query string.

@sbrunner, can you please confirm that this is what you had in mind?

elemoine:
Has this been tested?

rmichaelis:
Sorry. Still not tested. 
I do it today.

rmichaelis:
I think it's not working. 
After page reload, themes are loaded. 
Then I connect and the themes webservice is not called. 
Then disconnect and the themes webservice is still not called.  

capture

elemoine:
Argh.

I've only done tests in the dev tools' console:

```js
> $(document.body).scope().mainCtrl.roleId = 2;  // change the roleId in the main controller
> $(document.body).scope().$digest();  // trigger a digest
```

And this does trigger a new request to the `themes` service.

So the problem may be in the communication between the authentication directive and the main controller.

elemoine:
There was a typo in `index.html`. I had

```html
<app-authentication app-authentication-roleid="mainCtrl.rodeId"></app-authentication>
```

instead of 

```html
<app-authentication app-authentication-roleid="mainCtrl.roleId"></app-authentication>
```

It should work now. At least a request to the `themes` service should happen when the user logs in or logs out.

geoportailv3/static/js/authentication/authenticationdirective.js
```diff
@@ -207,17 +213,26 @@ app.AuthenticationController.prototype.hasError = function() {
...
+ * @param {string|undefined} mail Mail.
+ * @param {string|undefined} name Name.
  */
 app.AuthenticationController.prototype.setUserInfo = function(
-    login, role, role_id, mail, name) {
+    login, role, roleId, mail, name) {
   this['login'] = login;
   this['role'] = role;
-  this['role_id'] = role_id;
+  this['rodeId'] = roleId;
```
rmichaelis: 'roleId' and not 'rodeId'

elemoine: Woa! I really have a problem with this word. Thanks for catching this.

elemoine:
Thanks @rmichaelis. I fixed the remaining typo.

rmichaelis:
The functionality is ok.  
When I login and log out the theme is correctly reloaded, and the new layers are shown in catalog.

But the cache problem is still remaining. 
If a user has roleId = 1. 
Each response from the webservice using parameter roleId=1 is coming from the cache and not from the network.
That means, that the catalog is not updated with the last state of the database.

elemoine:
> But the cache problem is still remaining.
> If a user has roleId = 1.
> Each response from the webservice using parameter roleId=1 is coming from the cache and not from the network.
> That means, that the catalog is not updated with the last state of the database.

Can you confirm that the request to the `themes` web service includes a `role` parameter (whose value is the role id) when the user is logged in?

rmichaelis:
>Can you confirm that the request to the `themes` web service includes a `role` parameter (whose value is the role id) when the user is logged in?

Yes I confirm. It's working as expected. Except that only the first request of each roleId is coming from network. The next ones are coming from the cache.

elemoine:
> Yes I confirm. It's working as expected. Except that only the first request of each roleId is coming from network. The next ones are coming from the cache.

I understand. What cache are you talking about here?

rmichaelis:
>I understand. What cache are you talking about here?

I suspect the angular cache, because using the private mode of Firefox doesn't fix the problem.

elemoine:
What do you think about https://github.com/Geoportail-Luxembourg/geoportailv3/commit/f417d5538dbcdb519c406fc41e4d8c6cd7634918?

rmichaelis:
No, it doesn't work. Still the same behavior.

elemoine:
> No, it doesn't work. Still the same behavior.

That sucks. I will need to study this in more detail. Thanks a lot for testing!

rmichaelis:
It seems that responses are coming from BFCache
cache control is actually set to "max-age=864000, private"

elemoine:
> It seems that responses are coming from BFCache
> cache control is actually set to "max-age=864000, private"

@sbrunner, `themes` responses should not be cacheable, right?

jaykayone:
what about adding a random parameter?

sbrunner:
@elemoine , no it should, it's why we should add the role parameter to have one cache per role.
Then we can do many login / logout with only two request :-)

rmichaelis:
If the list of layers changed in webservice, how could the user have the last results ? 
Is a page reload enough ?

sbrunner:
Yes because we will also add a cache version :-)

jaykayone:
i have not understood your answer @sbrunner

elemoine:
> i have not understood your answer @sbrunner

@sbrunner is saying that, with https://github.com/camptocamp/c2cgeoportal/pull/1328, `themes` requests will include a `cache_version` parameter, whose value will change when the WSGI application cache is invalidated.

@rmichaelis, this also means that a page reload will be necessary to see changes in the database.

jaykayone:
OK for me

elemoine:
May I suggest to merge this and sort out the caching issues with separate GitHub issues/pull requests?

elemoine:
Do people agree with merging this?

rmichaelis:
Yes, I agree.

Issue #400, opened 2015-04-10
Add theme to permalink


elemoine:
As opposed to other states the name of the theme should be in the URL's path.

Restant c2c 01/04/2015 1j

elemoine:
@jaykayone could you please confirm that "theme" permalinks should look like as in version 2? E.g. http://eau.geoportail.lu/?zoom=2&X=100000&Y=80000.

elemoine:
http://map.geoportail.lu/eau?zoom=2&X=100000&Y=80000 also works. I guess this is the one the application should support, right?

elemoine:
New comment: in c2cgeoportal "theme" permalinks look like this: http://example.com/theme/eau. Note the word "theme" in the URL path. I guess this won't work for geoportailv3, right?

jaykayone:
eau.geoportail.lu is a rewrite to map.geoportail.lu/eau

the second one is the one to implementthe rest is done with rewrites

the /theme can be ok for us in v3 but with ?version=2 we have to support the /${theme}

elemoine:
Thanks Jeff. It makes sense. But we will need server-side changes to support `/{theme}` URLs. 

Another question: do we also store the current theme in the local storage? 

Thanks.

jaykayone:
yes, if someone comes through a rewritten subdomain it will be overloaded

elemoine:
> yes, if someone comes through a rewritten subdomain it will be overloaded

Ok.

Currently, if we find parameters in the URL, we do not look at the local storage. For example, if the URL is `/?zoom=15` then we use zoom 15 and default values for X and Y.

I wonder if the same should apply for themes. For example, if the URL is `/theme/eau` (or `/eau`) without any query parameters, should we use default values for X and Y, or should we use values from the local storage (if any)?

In other words, are themes handled differently from the other application states?

I am not sure my questions are clear. Maybe something to discuss at the next telco.

jaykayone:
this is a good question
I think it should be discussed at the next telco

jaykayone:
Decision: 
- No storage of the theme in the localstorage
- Different handling, which means that if only the theme is present we take the rest from localstorage

elemoine:
"question" tag removed.

pgiraud:
This morning, during the meeting, while we were talking about whether to get the theme from storage or not, I had a thought.
What about asking the user whether he/she wants to restore the previous context?
[...]

PR #404, opened 2015-04-13
Add theme to permalink

elemoine:


elemoine:
This PR requires https://github.com/Geoportail-Luxembourg/geoportailv3/pull/403, which should be merged first. But the code is ready for review, the relevant commit being 68e2e3e.

So permalinks of the forms `/theme/{theme}` and `/{theme}` are handled. When a permalink of the form `/{theme}` is received it is immediately changed to `/theme/{theme}`.

The behavior is as discussed in https://github.com/Geoportail-Luxembourg/geoportailv3/issues/400, that is:

* The theme is never stored in the local storage,
* When no theme is specified in the URL the "main" theme is used.

Please review.

There's one important bit that is missing, and that I'd like to discuss and address with a separate PR:

Currently URLs of the form `/{theme}` lead to 404s. This is because c2cgeoportal does not define a route for the `/{theme}` URL pattern. So we somehow need to add a route to handle `/{theme}` URLs. What do you think @sbrunner?

sbrunner:
In c2cgeoportal it's possible to add a route on `<apache_entry_point>/wsgi/<theme>` but for the `<apache_entry_point>/<theme>` we should create a rewrite rule, for that it can be possible but for that `c2c.template` should read the database. Then it's requires more development ...

jaykayone:
in geoportailv2, we had this type of config:

map.connect('/:theme', controller='home', action='index')

isn't this an option?

elemoine:
> In c2cgeoportal it's possible to add a route on <apache_entry_point>/wsgi/<theme> but for the <apache_entry_point>/<theme> we should create a rewrite rule, for that it can be possible but for that c2c.template should read the database. Then it's requires more development ...

The only thing we need is a route for `/<theme>` **in** the WSGI application. The rest will be addressed with rewrite rules on the Luxembourg side (@jaykayone?).

@sbrunner, for the `/<theme>` route, do you think it should be defined in c2cgeoportal or in the application? That sounds like an application-specific route to me. The problem is that the URL pattern is application-specific, but the corresponding view is defined in c2cgeportal.

elemoine:
> in geoportailv2, we had this type of config:
> 
> map.connect('/:theme', controller='home', action='index')
> 
> isn't this an option?

Yes, I am just wondering where the route should be defined.

sbrunner:
@elemoine with `/<theme>` you want to say `/wsgi/<theme>` right ?

jaykayone:
>  Yes, I am just wondering where the route should be defined.

@elemoine in that case it is not an apache/wsgi rewrite rule but a pyramid routing config, or am i missing something?

elemoine:
> @elemoine in that case it is not an apache/wsgi rewrite rule but a pyramid routing config

Yep.

elemoine:
> @elemoine with /<theme> you want to say /wsgi/<theme> right ?

Yes, this is why I said `/{theme}` **in** the WSGI app.

elemoine:
I rebased this. This is ready for review.

elemoine:
Does someone want to review this?

rmichaelis:
I tried to test with my ysgi instance : http://devv3.geoportail.lu/renaud
But I'm directly redirected to http://devv3.geoportail.lu/theme/main?version=3&zoom=10&X=693523&Y=6401189

Should I configure something ?

elemoine:
I hadn't thought about this problem. I'll see what I can do.

elemoine:
I changed the code. I hope it'll also work with modwsgi and `/<instanceid>/wsgi` type roots this time.

rmichaelis:
It works as expected now.

elemoine:
Thanks for testing again. Please tell me if you comments on the code and if/when I can merge this.

rmichaelis:
After reading the current thread, I'm not sure if that url should work or not ? http://devv3.geoportail.lu/renaud/theme/

rmichaelis:
Is the behavior with a a non existing theme correct or not ?
Should the user received a message or something telling him that the theme doesn't exist ?
http://devv3.geoportail.lu/renaud/theme/no_more_existing

elemoine:
> Is the behavior with a a non existing theme correct or not ?
> Should the user received a message or something telling him that the theme doesn't exist ?
> http://devv3.geoportail.lu/renaud/theme/no_more_existing

I assumed that this should never occur, but we can probably handle that if necessary.

rmichaelis:
Some themes are protected and need to be authenticated before being accessed.
We should be redirected to the login page.

elemoine:
> Some themes are protected and need to be authenticated before being accessed.
> We should be redirected to the login page.

I don't think that's related to that PR, no?

jaykayone:
ad: themes that do not exist: this is an error that should be caught

ad: authtification, this can be done in another PR but it is closely related

elemoine:
This PR is about reading/writing the "theme" part of permalinks. It's not related to authentication and redirecting the user to a login page. I'd suggest that we create a new issue and discuss redirection there.

But I agree we could better handle unknown themes. Should we just use the main theme when the one specified in the URL is unknown or should we also display an error message to the user? As I understand it permalinks are not something users manually create, so my feeling is that we can just use the "main" theme if the one specified in the URL is unknown, with no error message.

jaykayone:
yes this is ok for me too

elemoine:
I've changed the code to use the default theme (`main`) if the theme specified in the URL is not valid.

elemoine:
And I created https://github.com/Geoportail-Luxembourg/geoportailv3/issues/446 for the "protected themes" issue.

PR #407, opened 2015-04-14
Removing print from pgiraud's make config

pgiraud:


elemoine:
Please merge when the build passes on Travis.

PR #421, opened 2015-04-21
Rebuild template cache when directive partials change

elemoine:
This was removed in https://github.com/Geoportail-Luxembourg/geoportailv3/commit/c2ec0fa0d213e1afc86f487d5dc0572a9b74f7c5.

@sbrunner, this will need to be backported to c2cgeoportal.


elemoine:
Merging.

PR #425, opened 2015-04-22
Add theme rewrite rules

sbrunner:
Add a rewrite rule for all the themes, I get all the available themes but probably most of them should be removed...

Related to #404

### DIFF
--- a/apache/wsgi.conf.mako
+++ b/apache/wsgi.conf.mako
@@ -32,6 +32,10 @@ RewriteRule ^${apache_entry_point}admin/?$ /${instanceid}/wsgi/admin/ [PT]
 RewriteRule ^${apache_entry_point}search$ /${instanceid}/wsgi/fulltextsearch [PT]
 RewriteRule ^${apache_entry_point}s/(.*)$ /${instanceid}/wsgi/short/$1 [PT]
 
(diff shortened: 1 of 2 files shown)


jaykayone:
shouldn't the themes come directly from db instead of saving them in the yaml file?

sbrunner:
Why not but it should be made and takes some time probably 1-2 days

jaykayone:
/wsgi/theme/{theme} will always work?
It's only for /wsgi/{theme} that we need this, right?

In that case I can do all the rest with rewrites on the reverse proxy, so no need for further developments here

sbrunner:
yes, yes, exactly :-)

elemoine:
@jaykayone, yes, we thought it would be best to handle `/wsgi/{theme}` requests with an Apache rewrite rule. Please close this issue if the Lux team takes care of that.

jaykayone:
ok if this works http://devv3.geoportail.lu/main/wsgi/test (which is not the case today) we can take care of the rest 

but we have to be careful to support the old permalinks correctly

to be tested

jaykayone:
sorry I meant http://devv3.geoportail.lu/main/wsgi/theme/test

PR #427, opened 2015-04-24
Put the profile in the measures tools

pgiraud:
With this patch, the profile chart is located at the same place as the other measure tools.

This is how it looks:
screenshot from 2015-04-24 09 37 40


geoportailv3/static/js/measure/measuredirective.js
```diff
@@ -221,20 +222,22 @@ app.MeasureController = function($scope, $q, $http, ngeoDecorateInteraction,
...
-      this['profileOpen'] = false;
-    }
-  }, this));
+  goog.events.listen(measureProfile, ol.Object.getChangeEventType('active'),
+      /**
+       * @param {ol.ObjectEvent} evt Change active event.
+       */
+      function(evt) {
+        if (!measureProfile.getActive()) {
+          this['profileData'] = undefined;
```
elemoine: `profileData` should be initialized in the constructor:

```js
  /**
   * @type {…}
   */
  this['profileData'] = undefined;
```

geoportailv3/static/js/profile/profiledirective.js
```diff
@@ -226,6 +226,8 @@ app.ProfileController = function($scope) {
             p.getCoordinates().concat(newVal[i]['dist']));
       }
       this.line_ = lineString;
+    } else {
+      delete this.line_;
```
elemoine: I'd rather initialize `this.line_` to `null` in the constructor, and reset it to `null` here.

```js
  /**
   * @type {ol.geom.LineString}
   * @private
   */
  this.line_ = null;
```

elemoine:
I've just added two comments, regarding the initialization of the `profileData` and `line_` properties.

Issue #438, opened 2015-04-30
localstorage and private mode in safari


jaykayone:
when using private mode in safari, the application does not start because the accessing of the localstorage throws an uncaught error

elemoine:
So I guess should check that the local storage is available before attempting to store the application state into it.

jaykayone:
i think the problem is not only when storing but also when reading it..

elemoine:
> i think the problem is not only when storing but also when reading it..

Yes, true.

PR #439, opened 2015-05-05
Do not try to use localStorage if not available

elemoine:
With this commit we just don't use the localStorage if it's not available. In particular the localStorage is not available in Safari when the private mode is used.

@jaykayone, I couldn't fully test this commit as Chrome in incognito mode still allows to read/write from/to the localStorage. It's a simple change, so we could merge it and then have you test it again.

Fixes https://github.com/Geoportail-Luxembourg/geoportailv3/issues/438.


jaykayone:
ok for me

jaykayone:
OK works

elemoine:
Nice. Thanks for testing.

PR #448, opened 2015-05-16
Upgrade to GeoMapFish master

sbrunner:


CONST_Makefile
```diff
@@ -422,17 +428,17 @@ build-ngeo: \
...
-.build/venv/bin/db2pot: .build/requirements.timestamp
+$(VENV_BIN)/db2pot: .build/requirements.timestamp
 
 $(PACKAGE)/closure/%.py: $(CLOSURE_LIBRARY_PATH)/closure/bin/build/%.py
 	cp $< $@
 
-$(PACKAGE)/locale/$(PACKAGE)-js.pot: $(APP_HTML_FILES) $(APP_JS_FILES) $(APP_PARTIALS_FILES) .build/node_modules.timestamp
-	node tools/extract-messages.js $(APP_HTML_FILES) $(APP_JS_FILES) $(APP_PARTIALS_FILES) > $@
+$(PACKAGE)/locale/$(PACKAGE)-js.pot: $(APP_HTML_FILES) .build/node_modules.timestamp
+	node tools/extract-messages.js $(APP_HTML_FILES) > $@
```
sbrunner: Something missing in c2cgeoportal

CONST_Makefile
```diff
@@ -459,7 +465,6 @@ $(OUTPUT_DIR)/build.js: .build/node_modules.timestamp .build/build.js
 		bootstrap/dist/js/bootstrap.min.js \
 		proj4/dist/proj4.js \
 		d3/d3.min.js \
-		typeahead.js/dist/typeahead.bundle.min.js \
```
sbrunner: Missing in c2cgeoportal ?

CONST_Makefile
elemoine: Yes, I think so.

CONST_Makefile
elemoine: Yes. But if we add that to c2cgeoportal we should also add typehead as a node dependency to package.json. The problem is that package.json is in the create scaffold while CONST_Makefile is in the update scaffold.

elemoine:
@sbrunner do you need a review for this? Or is still WIP?

sbrunner:
Ready for review :-)

CONST_CHANGELOG.txt
```diff
@@ -100,26 +90,45 @@ Version 1.6
    Remove the no more needed `_query=url_params` argument of the
    `static_url` calls.
 
-8. In the `geoportailv3/templates/*.js` files do the following change:
```
rmichaelis: Point number 8 is missing ?

sbrunner: Effectively, thanks, I will fix in in c2cgeoportal :-)

PR #452, opened 2015-05-19
Lux interrogations

rmichaelis:
Work still in progress. 
Information comes from db query. Additional information can be added with python.
Information is displayed in the left side panel.

I'm working on queries on remote rest server.


rmichaelis:
Could you please review.

CONST_versions.txt
```diff
@@ -62,6 +62,7 @@ socketpool==0.5.3
 SQLAHelper==1.0
 SQLAlchemy==0.9.8
 Tempita==0.5.2
+suds>=0.4
```
sbrunner: You shouldn't modify this file, he will be overwritten on the c2cgeomapfish upgrade ...

You should add 
```
REQUIREMENTS += suds>=0.4
```
in the `geoportailv3.mk`

geoportailv3/PF.py
```diff
@@ -0,0 +1,416 @@
...
+import logging
+import os
+import sys
+# from geoadmin.model.meta import Session
+# from geoadmin.model.portail import UtiMesurageCommune
+
+
+# Classe pour acc?der aux attributs de la publicit? fonci?re##
+# Les classes import?es en haut ont ?t? g?n?r?es par wsdl2py
+# d?pendances: pyxml,fpconst,ZSI
```
elemoine: There are `?` instead of accentuated characters here. You may want to fix this.

geoportailv3/static/js/interrogation/interrogationdirective.js
```diff
@@ -0,0 +1,231 @@
+goog.provide('app.interrogationDirective');
+
+
+/**
+ * @param {string} appInterrogationTemplateUrl URL to directive template.
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+app.interrogationDirective = function(appInterrogationTemplateUrl) {
```
elemoine: I'd rather name this `queryDirective`, but I'll let you decide.

rmichaelis: You are right. QueryDirective is definitively better.

elemoine: There are lot of s/Interrogation/Query to make :-)

rmichaelis: Même pas peur :-)

rmichaelis: After renaming the directory, I had to make a cleanall to have it working in production mode.

geoportailv3/static/js/layerfactoryservice.js
```diff
@@ -96,7 +96,8 @@ app.getWmtsLayer_ = function(ngeoDecorateLayer) {
             '10', '11', '12', '13', '14', '15', '16', '17', '18', '19'
           ]
         }),
-        style: 'default'
+        style: 'default',
+        crossOrigin: 'anonymous'
```
elemoine: This shouldn't be in the diff, should it?

geoportailv3/static/js/layerfactoryservice.js
```diff
@@ -134,6 +135,7 @@ app.getWmsLayer_ = function(ngeoDecorateLayer) {
     var imageExt = app.getImageExtension_(imageType);
     var layer = new ol.layer.Image({
       source: new ol.source.ImageWMS({
+        crossOrigin: 'anonymous',
```
elemoine: Again, I'm surprised to see this in the diff.

geoportailv3/templates/default.html
```diff
@@ -0,0 +1,16 @@
...
+            <li ng-repeat="(key,data) in objects">
+            {{key}} : {{data}}
+            </li>
+        </ul>
+        <br><br><br>
+    </div>
+  </li>
+  </ul>
+  
+</div>
```
elemoine: Any reason for this Angular template to be placed in the `templates` directory?

rmichaelis: No this a mistake. It should not be here. I'll remove it.

geoportailv3/static/js/query/interrogationdirective.js
```diff
@@ -0,0 +1,234 @@
...
+    new ol.style.Style({
+      stroke: new ol.style.Stroke({
+        color: '#ffcc33',
+        width: 3
+      })
+    })
+  ];
+
+  var styles = {
+    'Point': [new ol.style.Style({
```
elemoine: I would be better to use `ol.geometry.GeometryType.POINT` here, for type-checking and minification reasons.

geoportailv3/static/js/layerfactoryservice.js
rmichaelis: Sorry but I don't understand why it shouldn't be in the diff ?

geoportailv3/static/js/query/interrogationdirective.js
```diff
@@ -0,0 +1,234 @@
...
+
+  /**
+   * The draw overlay
+   * @type {ol.FeatureOverlay}
+   * @private
+   */
+  this.overlay_ = new ol.FeatureOverlay({map: map,
+    style: function(feature, resolution) {
+      if (feature.getGeometry().getType() == 'Point' ||
+          feature.getGeometry().getType() == 'MultiPoint') {
```
elemoine: Again, I'd recommend using `ol.geometry.GeometryType.POINT`.

geoportailv3/static/js/query/interrogationdirective.js
```diff
@@ -0,0 +1,234 @@
...
+        width: 3
+      })
+    })
+  ];
+
+  var styles = {
+    'Point': [new ol.style.Style({
+      image: image
+    })],
+    'default': defaultStyle
```
elemoine: If your Angular templates do not include references to that property you do need to use quotes. Quotes prevent renaming, they should be used only when necessary.

elemoine: I think this is important to keep this in mind if we want to have a small build.

geoportailv3/static/js/query/interrogationdirective.js
```diff
@@ -0,0 +1,234 @@
...
+ * @export
+ * @ngInject
+ */
+app.QueryController = function($scope, $http,
+    appQueryTemplatesPath, getInfoServiceUrl,
+    getRemoteTemplateServiceUrl) {
+
+  this['content'] = [];
+  this.templatePath = appQueryTemplatesPath;
+  this.remoteTemplateUrl = getRemoteTemplateServiceUrl;
```
elemoine: Why not use `@private` variables?

geoportailv3/static/js/query/interrogationdirective.js
```diff
@@ -0,0 +1,234 @@
...
+      this.clearFeatures_();
+    }else {
+      this.highlightFeatures_(this.lastHighlightedFeatures_);
+    }
+  }, this));
+
+  var layers = map.getLayers().getArray();
+
+  var layersList = [];
+  var box = [];
```
elemoine: Any reason for these variables to be defined outside the `singleclick` listener?

geoportailv3/static/js/layerfactoryservice.js
elemoine: Because I thought it was part of a separate PR: https://github.com/Geoportail-Luxembourg/geoportailv3/pull/508/files.

geoportailv3/static/js/query/default.html
```diff
@@ -0,0 +1,9 @@
+<h1>{{layers['layerLabel'] | translate}}</h1>
```
elemoine: Any reason for using the bracket notation here? We always use the dot notation in templates.

geoportailv3/static/js/query/default.html
```diff
@@ -0,0 +1,9 @@
...
+<div ng-repeat="feature in layers['features']">
+    <table>
```
elemoine: We use two spaces indent elsewhere in Angular templates.

geoportailv3/static/js/query/interrogationdirective.js
```diff
@@ -0,0 +1,234 @@
...
+ * @param {string} appQueryTemplateUrl URL to directive template.
+ * @return {angular.Directive} The Directive Definition Object.
+ * @ngInject
+ */
+app.queryDirective = function(appQueryTemplateUrl) {
+  return {
+    restrict: 'E',
+    scope: {
+      'map': '=appQueryMap',
+      'selectedLayers': '=appQuerySelectedlayers',
```
elemoine: Is this used? Or should it be used?e

elemoine:
Done reviewing the JS code for now. I haven't reviewed the Python code, because I think most of it is a copy&paste (more or less) from the Pylons project.

geoportailv3/static/js/layerfactoryservice.js
rmichaelis: In fact,  "cors" pull request was merged into this one and not into master.

rmichaelis:
PR is updated with the corrections

geoportailv3/static/js/layerfactoryservice.js
elemoine: I see. I got confused because that's not something we usually do.

geoportailv3/static/js/query/querydirective.js
```diff
@@ -0,0 +1,242 @@
...
+};
+
+
+/**
+ * @param {Array<string>} features the features to highlight
+ * @private
+ */
+app.QueryController.prototype.highlightFeatures_ = function(features) {
+
+  var encOpt = {
```
elemoine: It would be safer to type this.

You can try to just give it a type first:

```js
/** @type {olx.format.ReadOptions} */
var encOpt = {
  dataProjection: 'EPSG:2169',
  featureProjection: this['map'].getView().getProjection()
};
```

If the compiler does not like it (because your object does not have a `rightHanded` property) then you'll have to use a type-cast:

```js
var encOpt = /** @type {olx.format.ReadOptions} */ ({
  dataProjection: 'EPSG:2169',
  featureProjection: this['map'].getView().getProjection()
});
```

Giving the object a type is safer because it ensures that the compiler will consistently rename object properties.

geoportailv3/static/js/layerfactoryservice.js
rmichaelis: I did it because both PR were related to each other.

geoportailv3/static/js/query/querydirective.js
```diff
@@ -0,0 +1,242 @@
...
+    dataProjection: 'EPSG:2169',
+    featureProjection: this['map'].getView().getProjection()
+  };
+
+  var jsonFeatures = (new ol.format.GeoJSON()).readFeatures({
+    'type': 'FeatureCollection',
+    'features': features},
+  encOpt
+  );
```
elemoine: I'd recommend using `readFeaturesFromObject` here:

```js
var jsonFeatures = (new ol.format.GeoJSON()).readFeaturesFromObject(
    /** @type {GeoJSONFeatureCollection} */ ({
      type: 'FeatureCollection',
      features: features
    }), encOpt);
```

And by using a type-cast for the GeoJSON object passed to `readFeaturesFromObject` you don't need to quotes the keys when defining that object.

geoportailv3/static/js/query/querydirective.js
```diff
@@ -0,0 +1,242 @@
...
+/**
+ * provides the template path according with the fact
+ * that the template for the current layer is remote or not
+ * @param {{remote_template: boolean, template: string, layer: string}} layer
+ * @return {string} the template path.
+ * @export
+ */
+app.QueryController.prototype.getTemplatePath = function(layer) {
+  if (layer['remote_template'] === true) {
+    return (this.remoteTemplateUrl_ + '?layer=' + layer['layer']);
```
elemoine: Minor, but brackets are useless here.

geoportailv3/static/js/query/querydirective.js
```diff
@@ -0,0 +1,242 @@
...
+  }
+  return (this.templatePath_ + '/' + layer['template']);
```
elemoine: Useless brackets.

geoportailv3/static/js/query/querydirective.js
```diff
@@ -0,0 +1,242 @@
...
+ * @param {string} getInfoServiceUrl
+ * @param {string} getRemoteTemplateServiceUrl
+ * @export
+ * @ngInject
+ */
+app.QueryController = function($scope, $http,
+    appQueryTemplatesPath, getInfoServiceUrl,
+    getRemoteTemplateServiceUrl) {
+
+  this['content'] = [];
```
elemoine: It would be good to provide a type for this. For example, it makes the code easier to understand to others.

geoportailv3/static/js/query/querydirective.js
```diff
@@ -0,0 +1,242 @@
...
+      })
+    })
+  ];
+
+  var styles = {
+    point: [new ol.style.Style({
+      image: image
+    })],
+    default: defaultStyle
+  };
```
elemoine: I would have defined a `@typedef` for that:

```js
/**
 * @typedef {{point: Array.<ol.style.Style>, default: Array.<ol.style.Style>}}
 */
app.QueryStyles;
```

geoportailv3/static/js/query/querydirective.js
```diff
@@ -0,0 +1,242 @@
...
+    return this['infoOpen'];
+  }, this), goog.bind(function(newVal, oldVal) {
+    if (newVal === false) {
+      this.clearFeatures_();
+    }else {
+      this.highlightFeatures_(this.lastHighlightedFeatures_);
+    }
+  }, this));
+
+  map.on('singleclick', goog.bind(function(evt) {
```
elemoine: To minimize strings in the code I'd recommend the enum objects defined in OpenLayers, and for [consistency](https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/js/locationcontrol.js#L70-L71) reasons I'd recommend using `goog.events.listen`:

```js
goog.events.listen(map, ol.MapBrowserEvent.EventType.SINGLECLICK, function(evt) {
  // …
}, false, this);
```

Note that you can pas `this` as the fifth argument and the creation of another function using `goog.bind`. This holds when using OpenLayers' `on` function as well.

elemoine:
I've added more comments. Please merge when they're addressed (and if you agree with them).

geoportailv3/static/js/query/querydirective.js
rmichaelis: I don't find the doc about `readFeaturesFromObject` . Do you know where I can find it ?

rmichaelis: I get this error when I use `readFeaturesFromObject` :  " Access to protected property readFeaturesFromObject of ol.format.GeoJSON not allowed here."

elemoine: > I don't find the doc about readFeaturesFromObject . Do you know where I can find it ?

It's [not](https://github.com/openlayers/ol3/blob/master/src/ol/format/geojsonformat.js#L425-L428) marked as `@api` in the OpenLayers 3 code, so it is not in the doc.

But we're different from other people :) Because we use the Closure Compiler, and compile the application and OpenLayers together. This means we have access to any OpenLayers symbol/function, including symbols that don't have the `@api` marker.

One could argue that using non-`@api` OpenLayers symbols is bad, because it means we use things that may disappear or change in the future. It's true, but the compiler helps us here. If something changes in OpenLayers the compiler will catch it immediately if we pay attention to correctly typing our code.

elemoine: > I get this error when I use readFeaturesFromObject : " Access to protected property readFeaturesFromObject of ol.format.GeoJSON not allowed here."

Hmm… ok that's because that function is marked as `@protected`. So yes, in that, we really can't use that function.

PR #458, opened 2015-05-19
To be valid a theme need not be displayed in switcher

elemoine:
Previously a theme that did not have `display_in_switcher` set could be considered valid. This PR changes that. To be valid the theme just needs to be present in the list of themes received from the server.

@rmichaelis, could you please review this?


jaykayone:
lgtm

PR #463, opened 2015-05-19
Adding a theme icon in the top bar

### LINKED ISSUE #454: Missing theme icon
An icon needs to be added next to logo for each theme.

pgiraud:
Fixes #454.

### DIFF
--- a/geoportailv3/static/less/icons.less
+++ b/geoportailv3/static/less/icons.less
@@ -26,6 +26,16 @@ button.icon {
   border: 0;
   text-indent: -9999px;
 }
+.theme-icon,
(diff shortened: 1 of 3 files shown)


elemoine:
Please merge.

jaykayone:
works great, but we have a problem with the placement of the icon relative to the gov logo.
we'll come back with a proposition

jaykayone:
can you make a propsition where the icon is shown at the left side of the search box?
what's more is that it does not seem to be vertically centered

Another idea we had was that a clic on the icon would open the theme switcher.
what do you think?

Issue #468, opened 2015-05-20
Support other print layouts than A4 and A3


elemoine:
For now we just print in A4 and A3. More layouts should be added.

PR #473, opened 2015-05-20
Add support for A2, A1, and A0 print-outs

elemoine:
Fixes https://github.com/Geoportail-Luxembourg/geoportailv3/issues/468.

Please review.


pgiraud:
Looks good. Please merge.

sbrunner:
All throes values should be get from the print capabilities ... not ?

elemoine:
Yes, eventually. At this point the ngeo print service [does not support](https://github.com/camptocamp/ngeo/blob/8babd0613aece4f2dd571a2ce7f8253dfa410987/src/services/print.js#L25) getting the print capabilities from MapFish Print. So the print configuration is duplicated.

sbrunner:
OK, thanks for the additional information :-)

Issue #475, opened 2015-05-20
Too many watchers…


elemoine:
With the "main" theme we currently have 1609 watchers. That's a lot!

By removing the catalog component from the page we're down to 137 watchers, so the catalog component sets up 1472 watchers by itself.

With https://github.com/Geoportail-Luxembourg/geoportailv3/pull/354 (Theme Switcher) we changed a number of expressions from "one-time binding" to "always binding". Most probably for good reasons… 

But I cannot observe problems when I re-add "one-time binding" expressions. See https://github.com/Geoportail-Luxembourg/geoportailv3/compare/elemoine_onetimebinding. With this branch I have 941 watchers instead of 1609, which would be a major drop.

@rmichaelis are you able to tell what won't work with my branch?

rmichaelis:
You are right, it seems to work as expected. 
I'm not able to reproduce the problem we had. The main categories and layers belonging to these are loaded without any problem.

Maybe a misunderstanding due to the old partials cache mechanism ??

jaykayone:
wasn't it an issue with translations when changing the language?

elemoine:
> wasn't it an issue with translations when changing the language?

I tested it, and could come up with a case where it did not work.

jaykayone:
and now ? :)

elemoine:
> and now ? :)

What do you mean?

jaykayone:
ah you added a "not" to your comment, then I can delete my answer

elemoine:
Yeah, sorry, I was too fast.

PR #479, opened 2015-05-20
One-time binding in the catalog tree node template

elemoine:
Use one-time binding again in the catalog tree node template, and reduce the number of watchers from 1609 to 941 (for the "main" theme).

Fixes https://github.com/Geoportail-Luxembourg/geoportailv3/issues/475.


pgiraud:
This can be merged. Thanks.

elemoine:
Ok let's merge it and see what problems come up :-)

Issue #480, opened 2015-05-20
Assertion failure when opening the "share" window


elemoine:
In debug mode I see an an assertion failure in the js console when I open the "share" window:

```
 AssertionError: Assertion failed
    at new goog.asserts.AssertionError (http://localhost:5000/node_modules/openlayers/node_modules/closure-util/.de…2b6477e735055b8f53284d5cf6a4af1b29da/closure/goog/asserts/asserts.js:62:20)
    at Object.goog.asserts.doAssertFailure_ (http://localhost:5000/node_modules/openlayers/node_modules/closure-util/.de…b6477e735055b8f53284d5cf6a4af1b29da/closure/goog/asserts/asserts.js:119:11)
    at Object.goog.asserts.assert [as assert] (http://localhost:5000/node_modules/openlayers/node_modules/closure-util/.de…b6477e735055b8f53284d5cf6a4af1b29da/closure/goog/asserts/asserts.js:149:18)
    at null.<anonymous> (http://localhost:5000/proj/dde378e4d7284809b87d092e723713b6/js/print/printdirective.js:156:20)
    at Scope.$get.Scope.$digest (http://localhost:5000/node_modules/angular/angular.js:14129:29)
    at Scope.$get.Scope.$apply (http://localhost:5000/node_modules/angular/angular.js:14391:24)
    at HTMLAnchorElement.<anonymous> (http://localhost:5000/node_modules/ngeo/src/directives/btngroup.js:130:19)
    at HTMLAnchorElement.jQuery.event.dispatch (http://localhost:5000/node_modules/jquery/dist/jquery.js:4670:9)
    at HTMLAnchorElement.jQuery.event.add.elemData.handle (http://localhost:5000/node_modules/jquery/dist/jquery.js:4338:46)
```

PR #483, opened 2015-05-20
Fix bug in print directive

elemoine:
The print directive shouldn't assume that the print mask is displayed when the "open" watcher reports that "open" is false. Indeed, we get "open is false" each time another panel is opened.

Fixes https://github.com/Geoportail-Luxembourg/geoportailv3/issues/480.


pgiraud:
Good catch.
Please merge.

PR #484, opened 2015-05-20
Adding a 'printing' notification for user

pgiraud:
Fixes partially #642.

### DIFF
--- a/geoportailv3/static/js/print/printdirective.js
+++ b/geoportailv3/static/js/print/printdirective.js
@@ -110,6 +110,11 @@ app.PrintController = function($scope, $timeout, ngeoCreatePrint,
   this['title'] = '';
 
   /**
+   * @type {boolean}
+   */
+  this['printing'] = false;
+
+  /**
    * @type {goog.events.Key}
    */
   var postcomposeListenerKey = null;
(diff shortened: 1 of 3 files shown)


geoportailv3/static/js/print/printdirective.js
```diff
@@ -110,6 +110,11 @@ app.PrintController = function($scope, $timeout, ngeoCreatePrint,
   this['title'] = '';
 
   /**
+   * @type {boolean|undefined}
+   */
+  this['printing'] = undefined;
```
elemoine: Why not false?

geoportailv3/static/js/print/printdirective.js
```diff
@@ -292,6 +298,7 @@ app.PrintController.prototype.handleGetStatusSuccess_ = function(ref, resp) {
   if (done) {
     // The report is ready. Open it by changing the window location.
     window.location.href = this.print_.getReportUrl(ref);
+    this['printing'] = false;
```
elemoine: Shouldn't you also reset `printing` to `false` in case of errors?

pgiraud: Indeed. Done.

pgiraud:
@elemoine Do you think we can merge it now?

elemoine:
We should! Thanks.

PR #493, opened 2015-05-21
Fix regression of theme switcher colors

### LINKED ISSUE #491: regression on theme switcher look
There is a regression 

in the theme switcher the text and icon colors is wrong now.
screen shot 2015-05-21 at 10 09 14
screen shot 2015-05-21 at 10 06 17

I am not sure how it was before but I'd say it was using light gray as text and icon color instead of a theme dependant one

pgiraud:
Fixes #491.


elemoine:
LGTM.

PR #495, opened 2015-05-21
Change scale to optimal when changing print layout

elemoine:
As discussed with @pgiraud this PR changes the print scale to the optimal scale when the user changes the print layout. This prevents the bad experience that occurs when selecting A0 and not seeing the print mask because it's outside the current viewport.


pgiraud:
+1
Please merge.

PR #501, opened 2015-05-27
Mobile specific layout for search input

pgiraud:
With this pull request I add the ability to do some search in devices with small screens (ie. mobile device).
When the user clicks on the _magnifier_ button, all the elements in the top bar are hidden, and the search input is displayed as well as a _cross_ button to get back to initial top bar.

This was a bit tricky and I'm not completely confident whether or not it will work on all devices.

There's also a usability issue for now since it would be good to have the keyboard being automatically shown when search input is displayed. Simply invoking `focus` on the input unfortunately doesn't seem to work on mobile devices (at least on safari iOS). I google a bit and tried several options without success.

Note : the list of results displays the same as in large screen devices.

img_4422

### DIFF
--- a/geoportailv3/static/js/search/search.html
+++ b/geoportailv3/static/js/search/search.html
@@ -3,7 +3,7 @@
 </button>
 <form class="navbar-form" role="search">
     <div class="form-group">
-        <input type="text" class="form-control" placeholder="{{'search_key_words'|translate}}"
+        <input type="text" autocorrect="off" autocapitalize="none" class="form-control" placeholder="{{'search_key_words'|translate}}"
             ngeo-search="ctrl.options" ngeo-search-datasets="ctrl.datasets" ngeo-search-listeners="ctrl.listeners">
             <span class="clear-button" ng-click="click($event)">&#10005;</span>
         </input>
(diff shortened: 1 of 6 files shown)


jaykayone:
works fine. the keyboard comes up after clicking into the field. that's ok for me

geoportailv3/static/js/search/search.html
```diff
@@ -1,4 +1,7 @@
-<form class="navbar-form pull-right hidden-xs" role="search">
+<button class="close-panel pull-right" ng-click="ctrl.mobileActive = false;" ng-show="ctrl.mobileActive">
```
elemoine: No need for semi-colon at end of expression.

geoportailv3/templates/index.html
```diff
@@ -54,7 +54,8 @@
             </ul>
           </li>
         </ul>
-        <app-search app-search-map="::mainCtrl.map"></app-search>
+        <app-search app-search-map="::mainCtrl.map"
+                    app-search-mobile-active="mainCtrl.mobileSearchActive"></app-search>
```
elemoine: We never use four-word attribute names elsewhere. So I'd use `app-search-mobileactive`.

petzlux:
Pierre, works great for me, just added a small commit

petzlux:
@pgiraud  Can this be merged ?

PR #500, opened 2015-05-26
Search for Coordinates

petzlux:
As requested by @jaykayone , users can enter LUREF coordinates in search field to go to specified location

fixes #492

PR #503, opened 2015-05-27
Fuzzy layer search

petzlux:
Add fuzzy search for layers and basemap layers.

This makes use of an external library [Fuse.js](https://github.com/krisk/Fuse) . I have written an externs file for the library.

### DIFF
--- a/CONST_Makefile
+++ b/CONST_Makefile
@@ -89,7 +89,7 @@ NGEO_LIBS_JS_FILES += \
 	node_modules/bootstrap/dist/js/bootstrap.min.js \
 	node_modules/proj4/dist/proj4.js \
 	node_modules/d3/d3.min.js \
-	node_modules/typeahead.js/dist/typeahead.bundle.min.js
+	node_modules/typeahead.js/dist/typeahead.bundle.min.js 
 
 # CGXP
 JSBUILD_MAIN_FILES = $(shell find $(PACKAGE)/static/lib/cgxp $(PACKAGE)/static/js -name "*.js" -print 2> /dev/null)
--- a/geoportailv3/externs/fuse.js
+++ b/geoportailv3/externs/fuse.js
@@ -1,0 +1,84 @@
+/**
+ * @fileoverview Externs for fuse.js
+ * @see https://github.com/krisk/Fuse/blob/master/README.md
+ * @externs
+ */
+
+/**
+ * @typedef {{
+ *    keys: Array.<string>,
+ *    id: (string|undefined),
+ *    threshold: (number|undefined),
+ *    distance: (number|undefined),
+ *    includeScore: (boolean|undefined)
+ * }}
+ */
+var FuseOptions;
+
+/**
+ * @type {Array.<string>}
+ */
+FuseOptions.keys;
+
+/**
+ * @type {string|undefined}
+ */
+FuseOptions.id;
+
+/**
+ * @type {number|undefined}
+ */
+FuseOptions.threshold;
+
+/**
+ * @type {number|undefined}
+ */
+FuseOptions.distance;
+
+/**
+ * @type {boolean|undefined}
+ */
+FuseOptions.includeScore;
+
+/**
+ * @typedef {{
+ *  item: (Object|string),
+ *  score: number
+ * }}
+ */
+var FuseResult;
+
+/**
+ * @type {Object|string}
+ */
+FuseResult.item;
+
+/**
+ * @type {number}
+ */
+FuseResult.score;
+
+/**
+ * @typedef (Array.<FuseResult>|Array.<string>)
+ */
+var FuseResults;
+
+/**
+ * @constructor
+ * @param {Array} list
+ * @param {FuseOptions} options
+ */
+function Fuse(list, options) {}
+
+/**
+ * @param {string} pattern
+ * @return {FuseResults} results
+ */
+Fuse.prototype.search = function(pattern) {};
+
+/**
+* Sets a new list for Fuse to match against.
+* @param {Array} list
+* @return {Array} The newly set list
+*/
+Fuse.prototype.set = function(list) {};
--- a/geoportailv3/static/js/search/searchdirective.js
+++ b/geoportailv3/static/js/search/searchdirective.js
@@ -33,6 +33,18 @@ app.BackgroundLayerSuggestion;
 
 
 /**
+ * @type {ol.layer.Layer}
+ */
+app.BackgroundLayerSuggestion.bgLayer;
+
+
+/**
+ * @type {string}
+ */
+app.BackgroundLayerSuggestion.translatedName;
+
+
+/**
  * @return {angular.Directive} The Directive Object Definition
  * @param {string} appSearchTemplateUrl
  * @ngInject
@@ -232,17 +244,29 @@ app.SearchDirectiveController = function($scope, $compile, gettextCatalog,
   var POIBloodhoundEngine = this.createAndInitPOIBloodhound_(
       ngeoCreateGeoJSONBloodhound, searchServiceUrl);
 
-  /** @type {Bloodhound} */
-  var LayerBloodhoundEngine = this.createAndInitLayerBloodhound_();
+  /** @type {Fuse} */
+  var layerFuseEngine =
+      new Fuse(this.layers_, {
+        keys: ['translatedName'],
+        threshold: 0.4,
+        distance: 100,
+        includeScore: true
+      });
 
-  /** @type {Bloodhound} */
-  var BackgroundLayerBloodhoundEngine = this.createAndInitLayerBloodhound_();
+  /** @type {Fuse} */
+  var backgroundLayerEngine =
+      new Fuse([], {
+        keys: ['translatedName'],
+        threshold: 0.4,
+        distance: 100,
+        includeScore: true
+      });
 
   $scope.$on('gettextLanguageChanged', goog.bind(function(evt) {
     this.createLocalAllLayerData_(
-        appThemes, LayerBloodhoundEngine, gettextCatalog);
+        appThemes, layerFuseEngine, gettextCatalog);
     this.createLocalBackgroundLayerData_(
-        appThemes, BackgroundLayerBloodhoundEngine, gettextCatalog);
+        appThemes, backgroundLayerEngine, gettextCatalog);
   }, this));
 
   goog.events.listen(appThemes, app.ThemesEventType.LOAD,
@@ -251,9 +275,9 @@ app.SearchDirectiveController = function($scope, $compile, gettextCatalog,
      */
       function(evt) {
         this.createLocalAllLayerData_(
-            appThemes, LayerBloodhoundEngine, gettextCatalog);
+            appThemes, layerFuseEngine, gettextCatalog);
         this.createLocalBackgroundLayerData_(
-            appThemes, BackgroundLayerBloodhoundEngine, gettextCatalog);
+            appThemes, backgroundLayerEngine, gettextCatalog);
       }, undefined, this);
 
   /** @type {TypeaheadOptions} */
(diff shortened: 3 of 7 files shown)


petzlux:
PS: based this of my work in #500  , will rebase with master once parent pull request is merged

geoportailv3/templates/index.html
```diff
@@ -196,6 +197,7 @@ <h2 translate>infos</h2>
          appModule.constant('printServiceUrl', "${request.route_url('printproxy')}");
          appModule.constant('shorturlServiceUrl', "${request.route_url('shortener_create')}");
          appModule.constant('defaultExtent', [425152.9429259216, 6324465.99999133, 914349.9239510496, 6507914.867875754]);
+         appModule.constant('maxExtent', [2.6, 47.7, 8.6, 51]);
```
elemoine: Have you tried to call ol.proj.transformExtent  here instead of calling it each time the max extent is needed?

geoportailv3/static/js/search/searchdirective.js
```diff
@@ -301,6 +365,64 @@ app.SearchDirectiveController =
...
+  for (var epsgCode in re) {
+    if ((m = re[epsgCode].regex.exec(searchString)) !== null) {
+      var easting = parseFloat(m[1]);
+      var northing = parseFloat(m[2]);
+      var point = /** @type {ol.geom.Point} */
+          (new ol.geom.Point([easting, northing])
+         .transform(epsgCode, 'EPSG:3857'));
+      if (ol.extent.containsCoordinate(
+          this.maxExtent_, point.getCoordinates())) {
+        var feature = /** @type {ol.Feature} */
```
elemoine: You don't need a type-cast here.

geoportailv3/static/js/search/searchdirective.js
```diff
@@ -301,6 +365,64 @@ app.SearchDirectiveController =
...
+            (new ol.Feature({
+              geometry: point,
+              'label': easting + 'E ' + northing + 'N ',
+              'epsgLabel': re[epsgCode].label
+            }));
```
elemoine: I'd recommend using `set` to avoid problems:

```
var feature = new ol.Feature(point);
feature.set('label', easting + 'E ' + northing + 'N ');
feature.set('epsgLabel', re[epsgCode].label);
```

See https://github.com/camptocamp/ngeo/blob/master/docs/developer-guide.md#custom-olobject-properties.

geoportailv3/externs/fuse.js
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @typedef {{
+ *    keys: (Array.<string>),
+ *    id: (string|undefined),
+ *    threshold: (number|undefined),
+ *    distance: (number|undefined),
+ *    includeScore: (boolean|undefined)
+ * }}
+ */
+var FuseOptions;
```
elemoine: I don't think a `@typedef` is sufficient. If these properties, or just some of them, are renamed by the compiler then it does mean a `@typedef` is not sufficient. In that case, this is what needs to be written:

```js
/**
 * @typedef {{
 *   keys: Array.<string>,
 *   ...
 * }}
 */
FuseOptions;

/**
 * @type {Array.<string>}
 */
FuseOptions.keys;

// same for every other properties…
```

geoportailv3/externs/fuse.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * @fileoverview Externs for fuse.js
+ * @see https://github.com/krisk/Fuse/blob/master/README.md
+ * @externs
+ */
+
+/**
+ * @typedef {{
+ *    keys: (Array.<string>),
```
elemoine: I don't think you need surrounding brackets here. They're only required when you use `|`, as for the other properties.

geoportailv3/static/js/search/searchdirective.js
```diff
@@ -301,6 +365,64 @@ app.SearchDirectiveController =
...
+ * @private
+ */
+app.SearchDirectiveController.prototype.matchLayers_ =
+    function(fuseEngine, searchString) {
+  var fuseResults = fuseEngine.search(searchString).slice(0, 5);
+  var results = [];
+  goog.array.forEach(fuseResults, function(element) {
+    results.push(element['item']);
+  });
+  return results;
```
elemoine: You could use `goog.array.map` instead:

```js
  return goog.array.map(fuseResults, function(r) {
    return r['item'];
  });
```

petzlux:
Okay will address these comments in parent pull request #500 , and after merge rebase

geoportailv3/templates/index.html
petzlux: for this to work, we would need to set this constant in (appmodule.js)[https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/js/appmodule.js]  ? 

I wanted to keep all constants in index, would make it easier to find stuff later ?

elemoine: > for this to work, we would need to set this constant in appmodule.js ?

Not sure, did you actually try to use `ol.proj.transformExtent` in `index.html`?

petzlux: Yes, didnt work if I did goog.require at the start of the function 
````
goog.require('ol.proj');
var appModule = angular.module('app');
...
````

Comment (elemoine) [inline geoportailv3/templates/index.html]: I did not mention using `goog.require('ol.proj')`.

Comment (petzlux) [inline geoportailv3/templates/index.html]: https://github.com/Geoportail-Luxembourg/geoportailv3/pull/500#issuecomment-105521694

Comment (petzlux) [inline geoportailv3/templates/index.html]: Okay, now realised I dont actually need goog.require('ol.proj') in index, works without it, will fix

Comment (elemoine) [inline geoportailv3/templates/index.html]: Yes, but in this comment the constant was not defined in `index.html`.

**Comment by sbrunner** [comment]
@elemoine this pull request add a new external library, do we as add it to c2cgeoportal, or is it a dependency that we will not have in c2cgeoportal and we should add a way do add it only in the project ?

**Comment by petzlux** [comment]
this is purely client side, so I dont think c2cgeoportal is concerned at all ... At the moment, its only for this project, I dont think its needed in ngeo ?

File: geoportailv3/externs/fuse.js (hunk 4 above)
Comment (petzlux) [inline geoportailv3/externs/fuse.js]: adressed here https://github.com/Geoportail-Luxembourg/geoportailv3/commit/1cf944548d0700e2e5f26205432504750d41a43e

File: geoportailv3/static/js/search/searchdirective.js [hunk 7]
```diff
@@ -340,6 +367,21 @@ app.SearchDirectiveController =
...
+ * @param {Fuse} fuseEngine
+ * @param {string} searchString
+ * @return {Array.<string>}
+ * @private
+ */
+app.SearchDirectiveController.prototype.matchLayers_ =
+    function(fuseEngine, searchString) {
+  var fuseResults = fuseEngine.search(searchString).slice(0, 5);
+  return goog.array.map(fuseResults, function(r) {
+    return r['item'];
```
elemoine: If you correctly type `r`, you should not need to use the bracket notation, as the type of `r` (`FuseResult`) is defined in an externs file.

I'd use something like this:

```js
  var fuseResults = /** @type Array.<FuseResult> */ (fuseEngine.search(searchString).slice(0, 5));
  return goog.array.map(fuseResults,
      /**
       * @param {FuseResult} r
       */
      function(r) {
        return r.item;
      });
```

package.json
```diff
@@ -25,6 +25,7 @@
     "typeahead.js": "~0.10.5",
     "temp": "~0.7.0",
     "walk": "~2.3.3",
-    "d3": "3.5.5"
+    "d3": "3.5.5",
+    "fuse": "krisk/Fuse#master"
```
elemoine: We should probably use the latest release: "krisk/Fuse#1.2.0".

petzlux: I depend on the master as a used function only has been added after the latest release

elemoine: In that case it would be safer to point to a specific commit.

geoportailv3/static/js/search/searchdirective.js
```diff
@@ -197,17 +198,29 @@ app.SearchDirectiveController =
...
 
-  /** @type {Bloodhound} */
-  var LayerBloodhoundEngine = this.createAndInitLayerBloodhound_();
-
-  /** @type {Bloodhound} */
-  var BackgroundLayerBloodhoundEngine = this.createAndInitLayerBloodhound_();
+  /** @type {Fuse} */
+  var LayerFuseEngine =
+      new Fuse(this.layers_, {
+        keys: ['translatedName'],
```
elemoine: You use a string here to refer to the `translatedName` property, while you use the dot notation elsewhere to access that property on a suggestion instance. That's where your problem is. I'll look at providing a solution to that problem…

geoportailv3/static/js/search/searchdirective.js
```diff
@@ -197,17 +198,29 @@ app.SearchDirectiveController =
   var POIBloodhoundEngine = this.createAndInitPOIBloodhound_(
       ngeoCreateGeoJSONBloodhound, searchServiceUrl);
 
-  /** @type {Bloodhound} */
-  var LayerBloodhoundEngine = this.createAndInitLayerBloodhound_();
-
-  /** @type {Bloodhound} */
-  var BackgroundLayerBloodhoundEngine = this.createAndInitLayerBloodhound_();
+  /** @type {Fuse} */
+  var LayerFuseEngine =
```
elemoine: This is minor, but `LayerFuseEngine` is not a type, so it's name should not start with a capital layer.

geoportailv3/static/js/search/searchdirective.js
elemoine: You may want to try using the `getFn` option:

```js
var layerFuseEngine = new Fuse(this.layers_, {
  keys: [''],
  getFn: function(item, keys) {
    var suggestion = /** @type {app.BackgroundLayerSuggestion} */ (item);
    return item.translatedName;
  },
  // …
});
```

This would require adding `getFn` to the externs file.

But, looking at your code, there are many variables that should typed but are not. The suggestion variables for example.

petzlux:
Merci @elemoine pour les infos et le mail, je vais regarder ca aujourd'hui et demain ...

CONST_Makefile
```diff
@@ -459,6 +459,7 @@ $(OUTPUT_DIR)/build.js: .build/node_modules.timestamp .build/build.js
 		bootstrap/dist/js/bootstrap.min.js \
 		proj4/dist/proj4.js \
 		d3/d3.min.js \
+		fuse.js/src/fuse.js \
```
elemoine: This file is not minifiied. I don't think we should add non-minified file to our build.

elemoine: I've just seen that the fuse repo does include a minified version of the library: `fuse.js/src/fuse.min.js`. This is the one we should add to our build.

geoportailv3/static/js/search/searchdirective.js
elemoine: @petzlux, did you at all try the suggestion provided above?

package.json
petzlux: I did a fork to also be able to use up to date minfied version

petzlux:
Works now!

petzlux:
@rmichaelis review please?

CONST_Makefile
```diff
@@ -89,7 +89,8 @@ NGEO_LIBS_JS_FILES += \
 	node_modules/bootstrap/dist/js/bootstrap.min.js \
 	node_modules/proj4/dist/proj4.js \
 	node_modules/d3/d3.min.js \
-	node_modules/typeahead.js/dist/typeahead.bundle.min.js
+	node_modules/typeahead.js/dist/typeahead.bundle.min.js \
+        node_modules/fuse.js/src/fuse.min.js
```
sbrunner: You shouldn't modify this file, You should add it in the geoportalv3.mk ...

petzlux: what not modify CONST_Makefile at all ? 

Can you please be more specific how to add an externs file to the build process through geoportailv3.mk ?

sbrunner: All the CONST* files will be overwritten on each update !

sbrunner: `NGEO_LIBS_JS_FILES += node_modules/fuse.js/src/fuse.min.js ` in the `geoportalv3.mk` file

rmichaelis:
lgtm

PR #508, opened 2015-06-02
cors header

rmichaelis:
Please review

### DIFF
--- a/geoportailv3/static/js/interrogation/interrogationdirective.js
+++ b/geoportailv3/static/js/interrogation/interrogationdirective.js
@@ -160,21 +160,26 @@ app.InterrogationController = function($scope, $http,
     }
   },this));
 
-  /*TODO : should implement CORS header before using it*/
-  /*
+
   map.on('pointermove', goog.bind(function(evt) {
     if (evt.dragging) {
       return;
     }
-
-
     var pixel = map.getEventPixel(evt.originalEvent);
     var hit = map.forEachLayerAtPixel(pixel, function(layer) {
-      return true;
+      if (goog.isDefAndNotNull(layer)) {
+        var metadata = layer.get('metadata');
+        if (goog.isDefAndNotNull(metadata)) {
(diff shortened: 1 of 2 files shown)


geoportailv3/static/js/interrogation/interrogationdirective.js
```diff
@@ -160,21 +160,26 @@ app.InterrogationController = function($scope, $http,
...
   map.on('pointermove', goog.bind(function(evt) {
     if (evt.dragging) {
       return;
     }
-
-
     var pixel = map.getEventPixel(evt.originalEvent);
     var hit = map.forEachLayerAtPixel(pixel, function(layer) {
-      return true;
+      if (goog.isDefAndNotNull(layer)) {
```
elemoine: Have you observed cases where `layer` was `undefined` or `null`?

geoportailv3/static/js/layerfactoryservice.js
```diff
@@ -134,6 +135,7 @@ app.getWmsLayer_ = function(ngeoDecorateLayer) {
     var imageExt = app.getImageExtension_(imageType);
     var layer = new ol.layer.Image({
       source: new ol.source.ImageWMS({
+        crossOrigin: 'anonymous',
```
elemoine: In most cases (and in production) WMS GetMap requests will be sent to the WMS proxy in the application. I hope that adding `crossOrigin: 'anonymous'` won't cause any problem in that case.

elemoine:
I just added two comments. Please merge when you think this is ready.

geoportailv3/static/js/interrogation/interrogationdirective.js
rmichaelis: Yes, it could happens when an overlay layer is existing. I'm not sure if it's normal or not but it happens.

elemoine: Ok, understood. I didn't know we used feature overlays already.

PR #515, opened 2015-06-04
Set print DPI 300 for nicer printed maps

### LINKED ISSUE #510: print quality
there is a problem with print quality.

print generated here: 

http://devv3.geoportail.lu/main/theme/main?version=3&zoom=15&X=680824&Y=6378491

Options: A4 portrait, 1:10000

elemoine:
Fixes #510.


jaykayone:
+1

Issue #489, opened 2015-05-21
print scales


jaykayone:
print scales should be the following:
1500, 2500, 5000, 10000, 15000, 20000, 25000, 50000, 80000, 100000, 125000, 200000, 250000, 400000

Please pay attention to the scale vs webmercator problem.

what's more is that in geoportalv2 the scales can vary according to the theme. (print_scales key in theme metadata)

elemoine:
> what's more is that in geoportalv2 the scales can vary according to the theme. (print_scales key in theme metadata)

Should this be implemented in v3 as well?

jaykayone:
yes please

elemoine:
See https://github.com/Geoportail-Luxembourg/geoportailv3/pull/516.

So with #516 we use the current theme's print scales, so I think your comment saying that "print scales should the following: …" doesn't hold anymore. Is that correct?

What about the "pay attention to the scale vs webmercator problem"? Could you please be more specific about that? Thanks.

jaykayone:
you are right for the first part, although we could take that array as fallback if no scales are defined in the theme metadata

the scale vs mercator problem is that due to the fact that webmercator meters are factor 1.5 to luref meters we choose to plot the scalebar in luref rather than in webmercator. the scale choosen for print should be a luref scale too and not a webmercator one

elemoine:
> you are right for the first part, although we could take that array as fallback if no scales are defined in the theme metadata

The themes that I get as an anonymous user all have print scales. Are there private themes with no print scales? I can add fallback print scales if you think it's necessary.

> the scale vs mercator problem is that due to the fact that webmercator meters are factor 1.5 to luref meters we choose to plot the scalebar in luref rather than in webmercator. the scale choosen for print should be a luref scale too and not a webmercator one

But do I need to do something about this? I mean: can I just use the print scales specified in the theme configuration?

jaykayone:
After the migration, they all have print-scales but I'd prefer to have a fallback to be sure.

Regarding the webmercator issue, I think that there is something to do, yes.
You certainly remember that when measuring distances we had to use spherical meters rather than webmercator meters to get correct distances. to have a correct scalebar, we have to specify another coordsys than webmercator.

In this case of the print I think that something has to be done to calculate the extent rectangle using spherical meters or in LUREF. I don't know how your code works in this case, I only wanted to stress this as it has been an issue everywhere else

elemoine:
[...]

PR #516, opened 2015-06-04
Use current theme's print scales

elemoine:
With this PR the print directive uses the current's print scales (`print_scales` in the theme's `metadata`) instead of using a fixed set of scales.

@jaykayone, could you please review this?

### DIFF
--- a/geoportailv3/static/js/print/printdirective.js
+++ b/geoportailv3/static/js/print/printdirective.js
@@ -216,6 +216,15 @@ app.PrintController.LAYOUTS_ = [
 
 
 /**
+ * @const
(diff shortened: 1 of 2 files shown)


jaykayone:
OK works as expected (see other remarks in https://github.com/Geoportail-Luxembourg/geoportailv3/issues/489)

elemoine:
If ok by you, I'll merge this and address the remaining issue with other PRs.

jaykayone:
OK

elemoine:
I've added a commit that adds default map print scales, which are used if the current theme does not have print scales.

PR #526, opened 2015-06-09
get proxy url from config file

### LINKED ISSUE #520: WMS proxy should not be hardcoded
Instead of hard coding the wms url, use a relative one.
If a config variable with a proxy url is existing then we should use it.

https://github.com/Geoportail-Luxembourg/geoportailv3/blob/8971ae9e44bf0a39a6c0a65b4222face877bd795/geoportailv3/static/js/layerfactoryservice.js#L134

rmichaelis:
fix #520 

Add the following entry in the local yaml file to use another proxy: 
```proxy_wms_url: http://devv3.geoportail.lu/main/wsgi/wms```

Could you please review.

### DIFF
--- a/geoportailv3.mk
+++ b/geoportailv3.mk
@@ -17,6 +17,7 @@ DISABLE_BUILD_RULES += test-packages test-packages-ngeo
 
 CONFIG_VARS += ldap
 CONFIG_VARS += sqlalchemy_engines
+CONFIG_VARS += proxy_wms_url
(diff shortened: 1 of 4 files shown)


elemoine:
Please merge when you think this is ready. And please tell us if you need to change our config files.

rmichaelis:
@elemoine @pgiraud  Please add this setting in your yaml file : 
```proxy_wms_url: http://devv3.geoportail.lu/main/wsgi/wms```

PR #543, opened 2015-06-11
Fix 486

rmichaelis:
Please review.


pgiraud:
I see one problem in my browser:
 - you can only hover the text of each item in the dropdown menu,
 - the background doesn't go to the right (only part of the width is highlighted).

screenshot from 2015-06-12 09 29 07

rmichaelis:
Yes you are right. 
There is the same issue in the infobar.

rmichaelis:
Is it better now ?

pgiraud:
It indeeds looks better.
Let's now wait for the travis builds to be fixed before merging.

Issue #451, opened 2015-05-19
Get browser language preferences 


petzlux:
Related to comment in #447 

Use https://github.com/camptocamp/ngeo/blob/master/src/services/getbrowserlanguage.js

jaykayone:
get language in the following order:

1) from URL
2) from localstorage
3) from browser pref

elemoine:
Precision: when the user selects a language using the language selector the language is added/updated in URL and in the local storage.

elemoine:
Another precision for 3) (from browser pref): we may want to read the browser pref language from the server, because the server already filters the language lists based on the languages it supports.

petzlux:
fixed by #545

PR #545, opened 2015-06-12
Permalink variables

petzlux:
adding missing permalink variables:

- [x] Background Layer
- [x] Layers
- [x] Layer Opacity
- [x] Language
- [x] Location information (with crosshair) see https://github.com/Geoportail-Luxembourg/geoportailv3/commit/f0ec17e5c3ab05858a072b61c4a2657dfc6451f7

### DIFF
--- a/geoportailv3/static/js/layerpermalinkservice.js
+++ b/geoportailv3/static/js/layerpermalinkservice.js
@@ -1,0 +1,254 @@
+/**
+ * @fileoverview This files defines an Angular Service for managing the
+ * selected layers permalinks
+ */
+
+goog.provide('app.LayerPermalinkManager');
+
+goog.require('app');
+goog.require('app.GetLayerForCatalogNode');
+goog.require('app.StateManager');
+goog.require('app.Themes');
+goog.require('ngeo.BackgroundLayerMgr');
+
+
+
+/**
+ * @constructor
+ * @param {app.StateManager} appStateManager
+ * @param {ngeo.BackgroundLayerMgr} ngeoBackgroundLayerMgr
+ * @param {app.GetLayerForCatalogNode} appGetLayerForCatalogNode
+ * @param {app.Themes} appThemes
+ * @ngInject
+ */
+app.LayerPermalinkManager =
+    function(appStateManager, ngeoBackgroundLayerMgr,
+        appGetLayerForCatalogNode, appThemes) {
+  /**
+   * @type {app.Themes}
+   * @private
+   */
+  this.appThemes_ = appThemes;
+
+  /**
+   * @type {app.StateManager}
+   * @private
+   */
+  this.stateManager_ = appStateManager;
+
+  /**
+   * @type {ngeo.BackgroundLayerMgr}
+   * @private
+   */
+  this.backgroundLayerMgr_ = ngeoBackgroundLayerMgr;
+
+  /**
+   * @type {app.GetLayerForCatalogNode}
+   * @private
+   */
+  this.getLayerFunc_ = appGetLayerForCatalogNode;
+};
+
+
+/**
+ * @param {Array.<ol.layer.Layer>} layers
+ * @private
+ */
+app.LayerPermalinkManager.prototype.setLayerState_ = function(layers) {
+  var layerIds = goog.array.map(layers, function(layer) {
+    return layer.get('queryable_id');
+  });
+  var opacities = goog.array.map(layers, function(layer) {
+    return layer.getOpacity();
+  });
+  var bgLayer = this.backgroundLayerMgr_.get(this.map_);
+  this.stateManager_.updateState({
+    'layers': layerIds.join('-'),
+    'opacities': opacities.join('-'),
+    'bgLayer': bgLayer.get('label')
+  });
+};
+
+
+/**
+ * @param {Array.<number>} layerIds
+ * @param {Array.<Object>} themes
+ * @private
+ */
+app.LayerPermalinkManager.prototype.applyLayerStateToMap_ =
+    function(layerIds, themes) {
+  var flatCatalogue = [];
+  for (var i = 0; i < themes.length; i++) {
+    var theme = themes[i];
+    goog.array.extend(flatCatalogue,
+        app.LayerPermalinkManager.getAllChildren_(theme.children)
+    );
+  }
+  goog.array.forEach(layerIds.reverse(),
+      function(layerId, layerIndex) {
+        var node = goog.array.find(flatCatalogue, function(catItem) {
+          return catItem.id === layerId;
+        });
+        var layer = this.getLayerFunc_(node);
+        // set opacity trough metadata to not interfere
+        // with the layer opacity manager service
+        layer.get('metadata')['start_opacity'] =
+            this.getStateValue_('opacities').reverse()[layerIndex];
+        this.map_.addLayer(layer);
+      }, this
+  );
+};
+
+
+/**
+ * @param {string} parameter
+ * @private
+ * @return {Array.<number>|undefined}
+ */
+app.LayerPermalinkManager.prototype.getStateValue_ = function(parameter) {
+  var result = '';
+  var response = this.stateManager_.getInitialValue(parameter);
+  if (goog.isDef(response) && response.length > 0) {
+    result = this.stateManager_.getInitialValue(parameter);
+  } else {
+    return undefined;
+  }
+  var items = [];
+  goog.array.forEach(result.split('-'), function(string) {
+    var value = parseFloat(string);
+    if (goog.isNumber(value)) {
+      items.push(value);
+    }
+  });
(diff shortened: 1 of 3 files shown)


petzlux:
please review @elemoine

petzlux:
fixes #451

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,254 @@
...
+                this.backgroundLayerMgr_.set(this.map_, layer);
+              }
+            }, this));
+
+        this.appThemes_.getThemesObject().then(
+            goog.bind(function(themes) {
+              if (goog.isDef(layerIds)) {
+                this.applyLayerStateToMap_(layerIds, themes);
+              }
+              this.setupWatchers_(selectedLayers);
```
elemoine: Do you set up new watchers each time themes are loaded?

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,254 @@
...
+};
+
+
+/**
+ * @param {Array.<ol.layer.Layer>} layers
+ * @private
+ */
+app.LayerPermalinkManager.prototype.setLayerState_ = function(layers) {
+  var layerIds = goog.array.map(layers, function(layer) {
+    return layer.get('queryable_id');
```
elemoine: We still don't have ids for layers? The only ids we have are the `queryable_id`s?

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,254 @@
...
+  goog.array.forEach(layerIds.reverse(),
+      function(layerId, layerIndex) {
+        var node = goog.array.find(flatCatalogue, function(catItem) {
+          return catItem.id === layerId;
+        });
+        var layer = this.getLayerFunc_(node);
+        // set opacity trough metadata to not interfere
+        // with the layer opacity manager service
+        layer.get('metadata')['start_opacity'] =
+            this.getStateValue_('opacities').reverse()[layerIndex];
```
elemoine: Do you need to call `getStateValue_('opacitites')` for each layer id?

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * @param {string} parameter
+ * @private
+ * @return {Array.<number>|undefined}
+ */
+app.LayerPermalinkManager.prototype.getStateValue_ = function(parameter) {
+  var result = '';
+  var response = this.stateManager_.getInitialValue(parameter);
+  if (goog.isDef(response) && response.length > 0) {
+    result = this.stateManager_.getInitialValue(parameter);
```
elemoine: Why do you call `getInitialValue` again?

geoportailv3/static/js/layerpermalinkservice.js
elemoine: Also, it is strange to call `reverse()` on something that may be `undefined`. (The return type of `getStateValue_` is `Array.<number>|undefined`.)

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,254 @@
...
+    var value = parseFloat(string);
+    if (goog.isNumber(value)) {
+      items.push(value);
+    }
+  });
+  if (items.length === 0) {
+    return undefined;
+  } else {
+    return items;
+  }
```
elemoine: You could use the ternary notation here:

```js
  return items.length === 0 ? undefined : items;
```

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,254 @@
...
+  // Add event listeners to existing selected layers
+  goog.array.forEach(layers.getArray(), function(layer) {
+    goog.events.listen(layer, ol.ObjectEventType.PROPERTYCHANGE,
+        goog.bind(function() {
+          this.setLayerState_(selectedLayers);
+        }, this));
+  }, this);
+  // Add event listener to all future selected layers
+  goog.events.listen(layers, ol.CollectionEventType.ADD,
+      goog.bind(
```
elemoine: You don't need to use `goog.bind` here as you pass `this` to `goog.events.listen` as the fifth argument (as done elsewhere).

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,254 @@
...
+  this.getLayerFunc_ = appGetLayerForCatalogNode;
+};
+
+
+/**
+ * @param {Array.<ol.layer.Layer>} layers
+ * @private
+ */
+app.LayerPermalinkManager.prototype.setLayerState_ = function(layers) {
+  var layerIds = goog.array.map(layers, function(layer) {
```
elemoine: `goog.require('goog.array')`

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,254 @@
...
+            /**
+             * @param {Array.<ol.layer.Base>} bgLayers
+             */
+            function(bgLayers) {
+              var stateLayerLabel =
+                  this.stateManager_.getInitialValue('bgLayer');
+              if (goog.isDefAndNotNull(stateLayerLabel)) {
+                var layer = /** @type {ol.layer.Base} */
+                    (goog.array.find(bgLayers, function(layer) {
+                  return layer.get('label') === stateLayerLabel;
```
elemoine: Isn't `label` a translated string? (not sure about this one)

geoportailv3/static/js/maincontroller.js
```diff
@@ -269,6 +289,40 @@ app.MainController.prototype.switchLanguage = function(lang) {
 
 
 /**
+ * @private
+ */
+app.MainController.prototype.initLanguage_ = function() {
+  this.scope_.$watchCollection(goog.bind(function() {
```
elemoine: `$watchCollection` for a scalar value? Are you sure?

geoportailv3/static/js/layerpermalinkservice.js
elemoine: Remember that themes are reloaded when the user logs in.

geoportailv3/static/js/maincontroller.js
petzlux: `$watch` probably better ?

geoportailv3/static/js/layerpermalinkservice.js
petzlux: `listenOnce` would help here ?

geoportailv3/static/js/layerpermalinkservice.js
petzlux: AFAIK yes ...

geoportailv3/static/js/layerpermalinkservice.js
elemoine: If you use `listenOnce` it won't work in the case where the user loads the application and then logs in, as you won't see the second `themes` response.

geoportailv3/static/js/maincontroller.js
elemoine: Does `$watchCollection` work at all in this case?

geoportailv3/static/js/layerpermalinkservice.js
petzlux: Okay, I think i will have to use the de registration function the watcher returns to destroy it each time themes get reloaded ...

geoportailv3/static/js/layerpermalinkservice.js
petzlux: so is the logic that we also need to add `goog.events` , `goog.bind`  ... to require statements ?

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -0,0 +1,259 @@
...
+  goog.events.listen(this.backgroundLayerMgr_, ngeo.BackgroundEventType.CHANGE,
+      function() {
+        var bgLayer = this.backgroundLayerMgr_.get(this.map_);
+        this.stateManager_.updateState({
+          'bgLayer': bgLayer.get('label')
+        });
+      }, undefined, this);
+
+  if (goog.isFunction(this.scopeWatcher_)) {
+    this.scopeWatcher_(); // destroy previous watcher
```
petzlux: @elemoine is this better now?

geoportailv3/static/js/layerpermalinkservice.js
petzlux: Yes but it doesnt get translated internally, just for display in the background layer switcher, which I dont touch ... so no problem

petzlux:
@elemoine  ready to merge now ?

Issue #467, opened 2015-05-20
Print legend support


elemoine:
Legends are not currently printed, but they should be.

sbrunner:
@elemoine actually is it possible to have a legend argument ?

If not is it possible to add a legend with the structure:
```
legend: [{name="<wiki_name>"}, ...]
```
?

elemoine:
Could you please clarify your questions? Are you talking about the print specifications?

sbrunner:
Yes, the arguments in the print spec.

elemoine:
The application can easily extend the print specifications object returned by ngeo's `createSpecs` function. Please tell me if that doesn't answer the question…

jaykayone:
Legends are now printed, but they are hard-coded

Issue #266, opened 2015-02-23
Print configuration


jaykayone:
Card for questions about print config

* [x] JRXML Files
* [x]  Legend implentation (HTML) cf #314
* [x] Scalebar in LUREF cf #313
* [x] WMTS layers don't work cf https://github.com/mapfish/mapfish-print/issues/208

Restant c2c 01/04/2015 : 2j
Restant c2c 01/05/2015 : 2j

jaykayone:
if I have several print layouts (a4 portrait, a4 landscape, a3 portrait, a3 landscape), I need one config defining them all and one jrml - file per layout. is that correct?

sbrunner:
yes :-)

jaykayone:
Our legends are html - documents that are embeded as iFrames.

I tried to use a HTML - Element of type iframe in the legend part

<body>
<iframe src='http://wiki.geoportail.lu/doku.php?id=fr:legend:tour_velo_national&do=export_html' width=400>
</body>

(test before starting to use parameters) but this does not seem to work.

@sbrunner: how would you implement this legend part? For the rest of the report everything is OK. I'll attach an example later on.

jaykayone:
OK, can't attach pdf, i'll send an email

jaykayone:
another issue I saw: the scalebar refers to distances measured in webmercator (as the data is in webmercator), but we would prefer it to refer to LUREF distances. is there a possibility?

elemoine:
> another issue I saw: the scalebar refers to distances measured in webmercator (as the data is in webmercator), but we would prefer it to refer to LUREF distances. is there a possibility?

Are you talking about the print report specifically here, or are you also referring to the ol3 map?

geoportallux:
I am talking about the print report, but the problem is the same for the map.
@petzlux is analyzing what this means for his infobar directive





Jeff Konnen
Chargé d’études – informaticien/Chef de projet

LE GOUVERNEMENT DU GRAND-DUCHÉ DE LUXEMBOURG
Ministère des Finances
Administration du cadastre et de la topographie
Service informatique & Service géoportail

54, avenue Gaston Diderich . L-1420 Luxembourg
Tél. (+352) 44901-261 . Fax (+352) 44901-333
E-mail: jeff.konnen@act.etat.lu

www.act.public.lu www.geoportail.lu



De : Éric Lemoine <notifications@github.com<mailto:notifications@github.com>>
[...]

PR #547, opened 2015-06-15
Render legend with weasyprint

### LINKED ISSUE #314: print legends

sbrunner:
fix #314, #467, #266 

part of #37


geoportailv3/static/js/print/printdirective.js
```diff
@@ -356,7 +356,9 @@ app.PrintController.prototype.print = function() {
           'scale': scale,
           'name': this['title'],
           'url': shorturl,
-          'qrimage': this.qrServiceUrl_ + '?url=' + shorturl
+          'qrimage': this.qrServiceUrl_ + '?url=' + shorturl,
+          'lang': "fr",
+          'legend': [{"name": "tour_velo_national"}, {"name": "tour_velo_national"}]
```
sbrunner: The above modification add attributes for test...

elemoine:
It looks good to me, but it'd be good to have @jaykayone or someone else from the Luxembourg team review this.

jaykayone:
First reaction: I had to install libffl-devel,libxml2-devel,libxslt-devel to make this compile.

jaykayone:
I cannot test further because of the ol3 - build problem

elemoine:
> First reaction: I had to install libffl-devel,libxml2-devel,libxslt-devel to make this compile.

So I think we should mention these dependencies in the README file.

sbrunner:
I don't understand the Travis error: https://travis-ci.org/Geoportail-Luxembourg/geoportailv3/builds/66839704 ...

sbrunner:
Dependencies added to the README :-)

sbrunner:
Just re-based for the Travis error :-)

geoportailv3/static/js/print/printdirective.js
```diff
@@ -359,7 +359,9 @@ app.PrintController.prototype.print = function() {
           'scale': scale,
           'name': this['title'],
           'url': shorturl,
-          'qrimage': this.qrServiceUrl_ + '?url=' + shorturl
+          'qrimage': this.qrServiceUrl_ + '?url=' + shorturl,
+          'lang': "fr",
```
elemoine: Why did you add `lang`?

print/print-apps/geoportailv3/a4_portrait.jrxml
```diff
@@ -12,8 +12,6 @@
 	<parameter name="mapSubReport" class="java.lang.String"/>
 	<parameter name="scalebarSubReport" class="java.lang.String"/>
 	<parameter name="name" class="java.lang.String"/>
-	<parameter name="legend" class="net.sf.jasperreports.engine.data.JRTableModelDataSource"/>
-	<parameter name="legendSubReport" class="java.lang.String"/>
```
elemoine: Why is a4 portrait the only file you've changed? Don't the other files deserve the same treatment?

jaykayone:
```
2015-06-15 15:40:27,836 ERROR [c2cgeoportal.views.proxy][Dummy-1] Error while processing request:
[Mon Jun 15 15:40:27.836710 2015] [:error] [pid 28730] java.lang.IllegalArgumentException: Expected a JSON Object as the value for the element with the path: 'spec.attributes.legend' but instead got a 'class org.json.JSONArray'.
[Mon Jun 15 15:40:27.836715 2015] [:error] [pid 28730] The value is: 
[Mon Jun 15 15:40:27.836719 2015] [:error] [pid 28730] [
[Mon Jun 15 15:40:27.836722 2015] [:error] [pid 28730]   {"name": "tour_velo_national"},
[Mon Jun 15 15:40:27.836726 2015] [:error] [pid 28730]   {"name": "tour_velo_national"}
[Mon Jun 15 15:40:27.836729 2015] [:error] [pid 28730] ]
[Mon Jun 15 15:40:27.836732 2015] [:error] [pid 28730] 
```

jaykayone:
I'll check again and keep you informed

print/print-apps/geoportailv3/a4_portrait.jrxml
sbrunner: Missing change in the others templates :-)

sbrunner: Just fixed :-)

geoportailv3/static/js/print/printdirective.js
sbrunner: We have one wiki page per language :-)

geoportailv3/__init__.py
```diff
@@ -28,6 +28,18 @@ def main(global_config, **settings):
...
+    # overwrite print routes
+    config.add_route(
+        "lux_printproxy_report_create",
+        "/printproxy/report.{format}",
+        request_method="POST"
+    )
+    config.add_route(
+        "lux_printproxy_report_get",
+        "/printproxy/report/{ref}",
+        request_method="GET"
```
elemoine: @sbrunner do you confirm that the cancel print route need not be overwritten?

sbrunner: Good question, probably we should overwrite it to remove the job...

elemoine: For your information, I am currently adding print cancel support to ngeo and the client side of geoportailv3, so I need a print service that accepts cancel requests.

sbrunner: This should already be the case ...

elemoine: I just tried to cancel a print request in the geoportailv3 project, and it does not work. I get a 404 from the mapfish print service:

```
2015-06-22 11:44:26,464 INFO  [c2cgeoportal.views.proxy][Dummy-4] Send query to URL:
http://localhost:8080/print-elemoine/print/geoportailv3/cancel/cd3020f7-a6e5-4665-a24d-b2767d001ef4@9a0c91e4-95ae-4495-9cf7-8378f802ab7f?.
2015-06-22 11:44:26,468 ERROR [c2cgeoportal.views.proxy][Dummy-4] Error 'Not Found' in response of URL:
http://localhost:8080/print-elemoine/print/geoportailv3/cancel/cd3020f7-a6e5-4665-a24d-b2767d001ef4@9a0c91e4-95ae-4495-9cf7-8378f802ab7f?.
2015-06-22 11:44:26,468 ERROR [c2cgeoportal.views.proxy][Dummy-4] Status: 404
2015-06-22 11:44:26,468 ERROR [c2cgeoportal.views.proxy][Dummy-4] Method: DELETE
2015-06-22 11:44:26,468 ERROR [c2cgeoportal.views.proxy][Dummy-4] --- With headers ---
Origin: http://localhost:5000
Accept-Language: en-US,en;q=0.8,fr;q=0.6
Accept-Encoding: gzip, deflate, sdch
Host: localhost:5000
Accept: application/json, text/plain, */*
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.81 Safari/537.36
Connection: keep-alive
Referer: http://localhost:5000/theme/main?version=3&zoom=15&X=680824&Y=6378491&debug&crosshair=false
Pragma: no-cache
Cache-Control: no-cache
2015-06-22 11:44:26,468 ERROR [c2cgeoportal.views.proxy][Dummy-4] --- Return content ---
2015-06-22 11:44:26,468 ERROR [c2cgeoportal.views.proxy][Dummy-4] <html><head><title>Apache Tomcat/7.0.52 (Ubuntu) - Error report</title><style><!--H1 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:22px;} H2 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:16px;} H3 {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;font-size:14px;} BODY {font-family:Tahoma,Arial,sans-serif;color:black;background-color:white;} B {font-family:Tahoma,Arial,sans-serif;color:white;background-color:#525D76;} P {font-family:Tahoma,Arial,sans-serif;background:white;color:black;font-size:12px;}A {color : black;}A.name {color : black;}HR {color : #525D76;}--></style> </head><body><h1>HTTP Status 404 - </h1><HR size="1" noshade="noshade"><p><b>type</b> Status report</p><p><b>message</b> <u></u></p><p><b>description</b> <u>The requested resource is not available.</u></p><HR size="1" noshade="noshade"><h3>Apache Tomcat/7.0.52 (Ubuntu)</h3></body></html>
```

sbrunner: OK, I will have a look :-)

sbrunner: It's a print issue ...

sbrunner: https://github.com/mapfish/mapfish-print/issues/260

sbrunner:
Cancel print route overwritten :-)

geoportailv3/views/luxprintproxy.py
```diff
@@ -0,0 +1,149 @@
...
+from c2cgeoportal.views.printproxy import PrintProxy
+from c2cgeoportal.lib.caching import init_cache_control, get_region
+
+log = logging.getLogger(__name__)
+cache_region = get_region()
+
+
+class LuxPrintProxy(PrintProxy):
+
+    ref_spec = {}
```
elemoine: I am not sure what this dict is for, but would that still work with multiple instances of application?

sbrunner: It's to be able to have the spec in the result message, and your right it's doesn't work with multiple wsgi instances ...
In the same idea the print needs a value repository and a queue to be able to have multi instance...

elemoine: > It's to be able to have the spec in the result message, and your right it's doesn't work with multiple wsgi instances ...

We need to verify this with the Lux team (@jakayone?), but that may well be a problem…



> In the same idea the print needs a value repository and a queue to be able to have multi instance...

Which is the case already? Or is there anything that needs to be done for this to be the case?

sbrunner: > Which is the case already?

No in the print I don't see anything to externalise it, but originally it was designed for this ...

elemoine: "externalize"?

sbrunner:
Yes, When we have only one print instance, we use an internal Java queue and a dictionary, but when we have multiple print instance we should have some external services that provide this, to gives to the client the same response on all the servers. (base on the same data)

elemoine:
> Yes, When we have only one print instance, we use an internal Java queue and a dictionary, but when we have multiple print instance we should have some external services that provide this, to gives to the client the same response on all the servers. (base on the same data)

I guess that before your PR the "multiple wsgi app instances + one mapfish print instance" case was covered. But now, with your PR, we cannot have several wsgi app instances. Is my understanding correct?

sbrunner:
An illustration for my previous comment
print

sbrunner:
@elemoine yes

elemoine:
In your diagram, print 1 and print 2 are mapfish print instances or print proxy instances?

sbrunner:
MapFish print :-)

jaykayone:
this could be a problem during development, but in production there will not be more than one instance per machine

sbrunner:
No it's just the inverse there isn't any issue on the development because there shouldn't by any interaction between the defendant instances, but on the production there is more than one instance of the print that should works together ...

elemoine:
> this could be a problem during development, but in production there will not be more than one instance per machine

The problem exists as soon as there is multiple instances of the application (and a load balancer to distribute the requests among these instances), be these instances on the same machine or on different machines.

jaykayone:
OK, in that case there is a problem, we will have multiple (min 2, probably 4) instances of the application running, probably each with it's own printserver, but not sure yet

sbrunner:
For the print is it relay necessary to have more than one dedicated print instance ?
Generally we don't like to have a tomcat and a wsgi application on the same server who is in a cluster ...
http://docs.camptocamp.net/c2cgeoportal/master/integrator/cluster.html

It's just a question but for the python webservice can't we use elastic search as a key-value registry ?

elemoine:
> It's just a question but for the python webservice can't we use elastic search as a key-value registry ?

I don't think I'd use ES for that. ES is a search engine, not a key/value store.

sbrunner:
I imagine that to don't add an other service ...

elemoine:
How did it work previously in geoportail v2?

sbrunner:
Sticky session ?

sbrunner:
Than we can use CouchDB ? It seem relay simple to create a cluster.

elemoine:
> Sticky session ?

I've just had a look at the geoportail v2 code, and it looks like printing is done synchronously. The print controller in the Pylons app does "create report" + "get report" + "merge legends into report" all in the same HTTP request/thread.

sbrunner:
Yes but it should have a sticky session to relay get the teport by the web client, It the console I see a `create.json` and a `[0-9]*.pdf` requests

jaykayone:
we also use cookies to force using the same backend between generating the print and grabbing it

sbrunner:
Effectively it's what we named Sticky session :-)

elemoine:
> Yes but it should have a sticky session to relay get the teport by the web client

I see.

So using sticky sessions would work here too. But I find it ugly personally… It would be ugly to use sticky sessions just because of the insertions of legends.

sbrunner:
+1 with @elemoine

elemoine:
@jaykayone, any existing memcached or Redis instance we could use? ;)

elemoine:
@sbrunner, we could possibly use a dedicated Postgres table, where the primary key would be the report reference identifier.

sbrunner:
Yes it should be possible, with memcached it's not so easy, Oliver dad many issue with it ...

elemoine:
> Yes it should be possible, with memcached it's not so easy, Oliver dad many issue with it ...

Many issue with what? Could you please be more specific?

elemoine:
> Many issue with what? Could you please be more specific?

Sorry. I had misread you. I now understood that he had issues with memcached.

jaykayone:
nope, no such thing here for now

elemoine:
I think the dedicated table option is the best at this point.

jaykayone:
can you help me understand what  would be stored in that table?

On Mon, Jun 22, 2015 at 6:18 PM, Éric Lemoine <notifications@github.com>
wrote:

> I think the dedicated table option is the best at this point.
>
> —
> <https://github.com/Geoportail-Luxembourg/geoportailv3/pull/547#issuecomment-114169001>
> .
>



-- 
Jeff Konnen

elemoine:
I can try. Look at @sbrunner's patch. It uses a class level dict to store the legend URLs between the "create report" request and the "get report" request. The keys of that dict are report reference identifiers generated by MapFish Print. So this won't work if the "create report" and "get report" requests land on different instances/machines because that dict is obviously not shared across instances. So we're looking for a shared place for this dictionary, and it sounds to me that at this point the database is our best shared location. We'd need a table with two columns: `ref`, the primary key, and a JSON column storing the corresponding legends information (URLs).

Tell us if you see a better way to solve that problem.

jaykayone:
ok understood, go for the db then

elemoine:
So I think we can merge this and address the remaining issue with a separate PR.

sbrunner:
OK, +1 for the db :-)

elemoine:
So please merge. I volunteer to look at adding the db table.

PR #549, opened 2015-06-15
no query info when measurement tools are opened

rmichaelis:
Please review

### DIFF
--- a/geoportailv3/static/js/query/querydirective.js
+++ b/geoportailv3/static/js/query/querydirective.js
@@ -22,7 +22,7 @@ app.queryDirective = function(appQueryTemplateUrl) {
       'map': '=appQueryMap',
       'infoOpen': '=appQueryOpen',
       'appSelector': '=appQueryAppselector',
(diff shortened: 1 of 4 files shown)


rmichaelis:
No comment about this one ?

geoportailv3/static/js/query/querydirective.js
```diff
@@ -21,7 +21,8 @@ app.queryDirective = function(appQueryTemplateUrl) {
     scope: {
       'map': '=appQueryMap',
       'infoOpen': '=appQueryOpen',
-      'appSelector': '=appQueryAppselector'
+      'appSelector': '=appQueryAppselector',
+      'measureOpen': '=appQueryMeasureopen'
```
elemoine: To me the query directive should not know about the measure component. Why not use `appQueryActive` as we do in other directives?

elemoine:
I just added a comment suggesting to change the attribute name to avoid coupling the query component to the measure component.

rmichaelis:
Yes you are right. 
This will be also used during the development of the drawing tools.

PR #574, opened 2015-06-18
clear highlight when a layer is added

rmichaelis:
Please review

### DIFF
--- a/geoportailv3/static/js/query/querydirective.js
+++ b/geoportailv3/static/js/query/querydirective.js
@@ -188,6 +188,17 @@ app.QueryController = function($timeout, $scope, $http,
     }
   }, this));
 
+  goog.events.listen(this.map_.getLayers(),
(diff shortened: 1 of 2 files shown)


rmichaelis:
No comment about this one ?

geoportailv3/static/js/query/querydirective.js
```diff
@@ -188,6 +188,17 @@ app.QueryController = function($timeout, $scope, $http,
...
 
+  goog.events.listen(this.map_.getLayers(),
+      ol.CollectionEventType.ADD,
+      /**
+       * @param {ol.CollectionEvent} e Collection event.
+       */
+      function(e) {
+        this['appSelector'] = undefined;
+        this['content'] = [];
+        this.clearFeatures_();
```
elemoine: Shouldn't there be a function that does these three operations?

PR #582, opened 2015-06-22
big tolerance for point small tolerance for others geometries

rmichaelis:
@jaykayone @petzlux Could you please check if the click tolerance is better now ?

http://devv3.geoportail.lu/renaud/theme/main?version=3&zoom=15&X=656759&Y=6407164&crosshair=true

### DIFF
--- a/geoportailv3/static/js/query/querydirective.js
+++ b/geoportailv3/static/js/query/querydirective.js
@@ -260,22 +260,35 @@ app.QueryController.prototype.singleclickEvent_ = function(evt) {
     }
   }
   if (layersList.length > 0) {
-    var buffer = 10;
-    var ll = ol.proj.transform(
+    var big_buffer = 20;
+    var small_buffer = 1;
+
+    var lb = ol.proj.transform(
         this.map_.getCoordinateFromPixel(
-        [evt.pixel[0] - buffer, evt.pixel[1] - buffer]),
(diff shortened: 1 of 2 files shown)


geoportailv3/static/js/query/querydirective.js
```diff
@@ -260,22 +260,35 @@ app.QueryController.prototype.singleclickEvent_ = function(evt) {
     }
   }
   if (layersList.length > 0) {
-    var buffer = 10;
-    var ll = ol.proj.transform(
+    var big_buffer = 20;
+    var small_buffer = 1;
```
elemoine: We don't use the snake notation elsewhere in the js code.

geoportallux:
works good on POI and parcels data.
It does not seem to be configured for remarkable trees though.

+ when clicking several times on the same spot or holding the mouse a little longer I get the location info on my laptop, which I should only get on a rightclick. can it be that I can generate a longpress event on my laptop?

rmichaelis:
Regarding the  remarkable trees, it seems that the ANF server does not respond (timed out) when it's called from my wsgi instance. 
Maybe a proxy config is missing ?

jaykayone:
check vars_main.conf

rmichaelis:
It works. Thanks. 
Remarkable trees layer is now returning results. But the remote anf server seems to have troubles with some spatial filters.

PR #598, opened 2015-06-24
use GET method with logout

### LINKED ISSUE #592: logout route
Hello @sbrunner ,

Logout does not work anymore in geoportailv3 since this . 
https://github.com/camptocamp/c2cgeoportal/blob/c8c34ec7b8a1fb0cfecfb0193a4e2f51b83b8a5a/c2cgeoportal/__init__.py#L486

Before modifying the code, I would like to know if there is a reason to limit the logout route to the GET method ?

rmichaelis:
fixes #592


elemoine:
LGTM.

sbrunner:
To me too :-)

PR #600, opened 2015-06-24
check if layer and opacities exists , and not duplicate, before adding

### LINKED ISSUE #590: Layer Permalink - duplication of permalinks if user login on reload
Need to check if layer id already in selectedLayers before writing to permalink to disable duplicates

petzlux:
fixes #590


rmichaelis:
LGTM

PR #607, opened 2015-06-25
uses only the visible layers

### LINKED ISSUE #605: getFeatureInfo - not on layers with opacity 0
there shouln't be any queries sent (and no mouse pointer changed) for layers with opacity 0

rmichaelis:
fixes #605


rmichaelis:
Please review.

PR #621, opened 2015-06-26
v2 permalink compatibility for background layers, layers and opacity

### LINKED ISSUE #555: Permalink backwards compatibility

petzlux:
fix #555


petzlux:
fixes #555

petzlux:
Ready for review

petzlux:
@rmichaelis can you review ?

rmichaelis:
V2 permalink does not contain the bgLayer parameter when the background is not modified. 
Because the default basemap is not the same in both V2 and V3, the result is different. 
It's a difference but probably not a problem.

jaykayone:
good remark, @rmichaelis 

None should translate to the default of the theme, which can be found in the metadata of the theme

petzlux:
@jaykayone I dont think that the theme bg_layer property is taken into account at the moment?

petzlux:
@rmichaelis good to merge otherwise ? we will address background layer in separate pull request, see #635

rmichaelis:
why are these 2 maps so different ? 
http://devv3rm.geoportail.lu/?zoom=4&X=75098.7415&Y=76009.28129&layers=addresses&layers_indices=6&layers_opacity=1&layers_visibility=true&bgLayer=webbasemap

http://map.geoportail.lu/?zoom=4&X=75098.7415&Y=76009.28129&layers=addresses&layers_indices=6&layers_opacity=1&layers_visibility=true&bgLayer=webbasemap

jaykayone:
because the zoomlevel lookup seems to be one off

petzlux:
No its more subtle than that ... the zoomlevels are not equivalent ...

jaykayone:
yes that's for sure but in the particular case, the smaller one would be a much better match

petzlux:
i will try and find a compromise

rmichaelis:
but in this case, scales are the same ?

petzlux:
@rmichaelis @jaykayone can you check if the new zoom level conversion is okay

jaykayone:
bgOpacity=0 should switch to ortho background

jaykayone:
zoomlevels are perfect.

in some thematic portals (like go), there are additional levels (11+). we should at least have a fallback to 19 for those

petzlux:
@jaykayone added your feature requests.

petzlux:
Ready for review

jaykayone:
works fine

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -270,8 +288,71 @@ app.LayerPermalinkManager.prototype.init =
 
         this.appThemes_.getThemesObject().then(
             goog.bind(function(themes) {
-              if (goog.isDef(layerIds)) {
-                this.applyLayerStateToMap_(layerIds, themes);
+              var flatCatalogue = [];
+              for (var i = 0; i < themes.length; i++) {
+                var theme = themes[i];
+                goog.array.extend(flatCatalogue,
```
elemoine: `goog.require('goog.array')` missing.

geoportailv3/static/js/layerpermalinkservice.js
```diff
@@ -270,8 +288,71 @@ app.LayerPermalinkManager.prototype.init =
...
+              var layerIds = [];
+              /**
+               * @type {Array.<number>|undefined}
+               */
+              var opacities = [];
+
+              if (this.initialVersion_ === 2) {
+                var layerString = this.stateManager_.getInitialValue('layers');
+                if (goog.isDefAndNotNull(layerString) &&
+                    !goog.string.isEmpty(layerString)) {
```
elemoine: `goog.require('goog.string')` missing.

geoportailv3/static/js/layerpermalinkservice.js
petzlux: already there

geoportailv3/static/js/layerpermalinkservice.js
petzlux: will add in new pull request

PR #640, opened 2015-07-01
Store the print spec in the database

sbrunner:


geoportailv3/models.py
```diff
@@ -77,3 +77,13 @@ class LuxGetfeatureDefinition(Base):
...
     geometry_column = Column(Unicode, label=_(u'Geometry column name'))
+
+
+class LuxPrintJob(Base):
+    __tablename__ = 'lux_print_job'
+    __table_args__ = {'schema': _schema}
+
+    id = Column(Integer, primary_key=True)
+    ref = Column(String, index=True, unique=True, nullable=False)
+    spec = Column(Unicode)
```
elemoine: I thought you'd use a JSON column.

geoportailv3/models.py
```diff
@@ -77,3 +77,13 @@ class LuxGetfeatureDefinition(Base):
...
+    creation = Column(DateTime)
```
elemoine: What is this column for?

geoportailv3/models.py
```diff
@@ -77,3 +77,13 @@ class LuxGetfeatureDefinition(Base):
...
     poi_id_collection = Column(Unicode, label=_(u'Id of the poi collection'))
     geometry_column = Column(Unicode, label=_(u'Geometry column name'))
+
+
+class LuxPrintJob(Base):
+    __tablename__ = 'lux_print_job'
+    __table_args__ = {'schema': _schema}
+
+    id = Column(Integer, primary_key=True)
+    ref = Column(String, index=True, unique=True, nullable=False)
```
elemoine: Why not use `ref` as the primary key? I don't think we need a database sequence in this case.

geoportailv3/views/luxprintproxy.py
```diff
@@ -83,8 +85,9 @@ def lux_report_create(self):
 
     @view_config(route_name="lux_printproxy_report_cancel")
     def lux_cancel(self):
-        ref = self.request.matchdict.get("ref")
-        del self.ref_spec[ref]
+        DBSession.query(LuxPrintJob).filter(
+            LuxPrintJob.ref == self.request.matchdict.get("ref")
+        ).delete()
```
elemoine: You'd just use `DBSession.query(LuxPrintJob).get(ref).delete()` if `ref` was the primary key.

elemoine: Which would be very efficient…

geoportailv3/models.py
sbrunner: Get the creation date, can be useful if we want to clean-up old jobs...

geoportailv3/models.py
sbrunner: Good question I hesitate to do that than I will do it :-)

geoportailv3/models.py
sbrunner: I never use JSON column  type, then I just have some quick questions:
What the advantages? Type verification? Shouldn't we passed the JSON on the read and on the write than it will takes some time?

geoportailv3/views/luxprintproxy.py
sbrunner: > Which would be very efficient…

For witch reason? for me it should be the same about efficiency...

sbrunner:
Thanks @elemoine for the review :-)

geoportailv3/views/luxprintproxy.py
pgiraud: I don't know if it's more efficient or not. But I find it a bit more elegant. ;-)

geoportailv3/models.py
elemoine: True.

geoportailv3/views/luxprintproxy.py
elemoine: I assume it's more efficient because Postgres uses a hashtable whose keys are the primary keys.

elemoine: And we just don't need a sequence in that case.

geoportailv3/models.py
elemoine: No real advantage in this case I guess. It just made more sense to me as the original data that we want to store is JSON.

geoportailv3/views/luxprintproxy.py
sbrunner: I also index the ref column than I don't think that it will have a performance change, I wool do the change only to be more elegant :-)

geoportailv3/models.py
sbrunner: I don't see Json as en sqlalchemy type ...

LUX_alembic/versions/34e77893d2fa_add_print_spec_table.py
```diff
@@ -0,0 +1,33 @@
...
+from alembic import op, context
+from sqlalchemy import Column
+from sqlalchemy.types import String, Unicode, DateTime
+
+
+def upgrade():
+    schema = context.get_context().config.get_main_option("schema")
+    op.create_table(
+        "lux_print_job",
+        Column("id", String(100), primary_key=True),
```
elemoine: Are we sure about the `100` limit?

elemoine:
I just added another minor comment. Otherwise looks good to me.

LUX_alembic/versions/34e77893d2fa_add_print_spec_table.py
sbrunner: yes

sbrunner: It's a little bit less, by mind about 80

elemoine: @sbrunner, any answer for me? I am just curious about this.

PR #647, opened 2015-07-02
Fix location info issues

### LINKED ISSUE #643: Crosshair is to true too often
If you connect to application: http://devv3.geoportail.lu/main/wsgi/theme/main
And then open the info panel by clicking on the info button in the bottom left bar. A "crosshair=true" parameter is added to the url. If you copy this url and paste it in a new tab, you'll see a marker in the map.

Moreover, if you right click somewhere in the map, the x and y coordinates in the url are for the clicked point (not the center of the map). Which is probably expected. But if you then close the info panel, x and y don't change back to the center of the map. If you share the link (using the share tool), the link doesn't show the same view.

petzlux:
fix #643

### DIFF
--- a/geoportailv3/static/js/locationinfo/locationinfodirective.js
+++ b/geoportailv3/static/js/locationinfo/locationinfodirective.js
@@ -111,6 +111,13 @@ app.LocationinfoController =
     }
     if (newVal === false) {
       this.stateManager_.updateState({'crosshair': false});
+      var mapCenterCoordinate = this['map'].getView().getCenter();
+      if (goog.isDefAndNotNull(mapCenterCoordinate)) {
+        this.stateManager_.updateState({
+          'X': parseInt(mapCenterCoordinate[0], 0),
+          'Y': parseInt(mapCenterCoordinate[1], 0)
+        });
+      }
       this['appSelector'] = undefined;
       this.coordinate_ = null;
       this.vectorOverlay_.clear();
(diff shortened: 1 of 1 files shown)


petzlux:
@pgiraud  please review?

jaykayone:
I rebased this, travis should now be OK....

geoportailv3/static/js/locationinfo/locationinfodirective.js
```diff
@@ -111,10 +111,16 @@ app.LocationinfoController =
     }
     if (newVal === false) {
       this.stateManager_.updateState({'crosshair': false});
+      var mapCenterCoordinate = this['map'].getView().getCenter();
+      if (goog.isDefAndNotNull(mapCenterCoordinate)) {
```
pgiraud: Is there any case where it can be null or undefined?

pgiraud:
Sorry if I'm late reviewing this pull request but there's something I don't get with your code.
It looks like the main interest of the `coordinate_` property is to put a watcher on it and when it's not null set a new value to `location`. Why not modifying the `location` directly? (ie. in `showInfoPane_`)
I may be wrong but the code could be simplified a bit.

geoportailv3/static/js/locationinfo/locationinfodirective.js
elemoine: `getCenter`'s return type is `ol.Coordinate|undefined`. But given the way we use the View in the application `getCenter` should never return `null` or `undefined`. An assertion is therefore the best thing to use here.

```js
goog.asserts.assert(goog.isDefAndNotNull(viewCenter));
```

petzlux:
@pgiraud , Regarding the coordinate_ variable, you were right, and I removed the variable.

Can you please check this branch on your machine, I have now issue after a rebase from master, that there is now a distortion of the map if the location info window gets closed. Before, as you know the inverse was the case, but you fixed that.

PR #650, opened 2015-07-02
open info window only if results close info window if no result

### LINKED ISSUE #648: Click on map should only open Info window if result returned

rmichaelis:
fixes #648


petzlux:
LGTM

PR #652, opened 2015-07-02
Less travis faild by trying two times the download tasks

sbrunner:


rmichaelis:
good idea :-)

elemoine:
The build failed :(

sbrunner:
Because of that:  #653 then I will rebase :-)

sbrunner:
Tests will pass ... :-)

jaykayone:
they did, please merge

petzlux:
can we merge this so that hopefully travis will function better ?

elemoine:
I think we can merge this, but I am afraid this would hide a problem with pypi.camptocamp.net.

I just tried to update my instance and it failed because of errors when downloading packages from pypi.camptocamp.net.

@sbrunner, have you also experienced problems again with our pypi service?

jaykayone:
let's merge this anyway

PR #653, opened 2015-07-02
Upgrading angular to 1.4

pgiraud:


pgiraud:
This pull request finally passes tests. Please review.

sbrunner:
I relay don't know about the measure directive, otherwise that's looks good :-)

pgiraud:
The modification in measuredirective only means that be null. Without that closure compiler complains.

elemoine:
Looks good to me too.

PR #654, opened 2015-07-02
Pgiraud crosshair map distortion

pgiraud:
#653 needs to be merged first.

I had a lot of troubles fixing this one with the help of @elemoine.
To summarize, we need to wait for ngAnimate to be "ready" before we can put observers on animations.
But we also had troubles with weird behavior in Angular which we haven't fully understood.

### DIFF
--- a/geoportailv3/static/js/locationinfo/locationinfodirective.js
+++ b/geoportailv3/static/js/locationinfo/locationinfodirective.js
@@ -42,6 +42,7 @@ app.module.directive('appLocationinfo', app.locationinfoDirective);
  * @constructor
  * @param {angular.Scope} $scope
  * @param {angular.$timeout} $timeout
+ * @param {angular.$animate} $animate Angular animate service.
  * @param {app.GetShorturl} appGetShorturl
  * @param {app.GetElevation} appGetElevation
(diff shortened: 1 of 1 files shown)


pgiraud:
Not ready for review yet.

geoportailv3/static/js/locationinfo/locationinfodirective.js
```diff
@@ -298,6 +313,17 @@ app.LocationinfoController.prototype.showInfoPane_ =
         this['qrUrl'] = this.qrServiceUrl_ + '?url=' + shorturl;
       }, this));
 
+  // we need to wait for angular to be ready to manage animations
+  var dereg = this.scope_.$watch(goog.bind(function() {
+    return this.$animate_.enabled();
+  }, this), goog.bind(function(newVal) {
+    if (newVal) {
+      this.$timeout_(goog.bind(function() {
```
elemoine: I think the use of `$timeout` here deserves a comment in the code, even if why it's needed is not fully understood. We should be honest to ourselves :)

pgiraud:
Comment added. Can I merge?

elemoine:
Yes.

PR #672, opened 2015-07-07
go back to 96 dpi resolution in print 

jaykayone:
to avoid symbol scaling problems. a better solution will be implemented later on

### DIFF
--- a/geoportailv3/static/js/print/printdirective.js
+++ b/geoportailv3/static/js/print/printdirective.js
@@ -294,7 +294,7 @@ app.PrintController.MAP_SIZES_ = [
  * @type {number}
  * @private
  */
-app.PrintController.DPI_ = 300;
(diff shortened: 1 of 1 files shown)


elemoine:
Looks good to me.

But why 96? Just curious…

jaykayone:
because ESRI seems to take 96 as the standard resolution in their WMS implementation

elemoine:
Really! The OGC standard is ~90 DPI (25.4 / 0.28).

jaykayone:
i have not found it written anywhere, but all their document cite 96 as an example when taling about printing and mapservices.
esri-specific protocols (like exportMap) accept a dpi setting, but not wks. so I assumed they were using 96 and the result seems correct

elemoine:
Ok thank you for the explanation. We can talk about this again during milestone 2.

PR #680, opened 2015-07-07
Update of translations

petzlux:


jaykayone:
don't merge, there are wrong translations

petzlux:
@jaykayone better ?

jaykayone:
a little bit

petzlux:
@pgiraud @elemoine  

The translations for the measure tools tooltips are not taken into account. They are present in the .po files, but missing from the generated geoportailv3.json language files and thus the tooltips are always in english.

elemoine:
> The translations for the measure tools tooltips are not taken into account. They are present in the .po files, but missing from the generated geoportailv3.json language files and thus the tooltips are always in english. 

@petzlux did you remove that comment?

jaykayone:
good question ;)

petzlux:
Yes I did, somehow a rebuild did the trick , probably my stupidity rather than a real problem.

jaykayone:
let's merge

Issue #664, opened 2015-07-06
map distortion


jaykayone:
when using the getfeatureinfo, the map is now distorted

jaykayone:
this is a regression

jaykayone:
as @pgiraud is not there today, can you have a look please @elemoine?

elemoine:
I did not have time for that. I just noticed that the problem is related to https://github.com/camptocamp/ngeo/pull/256. In the case of the "info panel" the animation callback is called with a target element which is not the one onto which the resizemap directive is attached.

jaykayone:
@elemoine do you have the time to work on this today? 
If not, do you think, we can fix it on the application side, or does this have to be tackled in ngeo?

It's the last issue preventing us to make user tests.

Thanks
Jeff

pgiraud:
I'm back to work. I'll dig into this this afternoon.

jaykayone:
thanks

pgiraud:
It looks like it happens in other cases. Here's an other one:
 - right click on map -> info panel opens,
 - click on closing cross -> info panel closes and map is distorted.

jaykayone:
yes, that's correct

rmichaelis:
it seems to happen as soon as the left panel is not opened by one of the 3 left button.
I have the feeling that it happens when the side panel is opened through a binding in a directive.

pgiraud:
It looks like something weird happens when there's a ngRepeat or ngShow in child elements.
I made a sample to show this. http://plnkr.co/edit/KEIfoHDiw7n97ljLm2CZ
And sent a message to the mailing list https://groups.google.com/forum/#!topic/angular/T0j0gg1G2P0

jaykayone:
Is there an alternative way of repairing this? 
if possible I do not want to wait for angular to fix a bug

pgiraud:
I understand. I'm trying to find a workaround.
By the way, here's an example showing how easy it can be reproduced.
https://github.com/camptocamp/ngeo/commit/41a0753c2cb80e97339956fd0781b423f78dc1e1

rmichaelis:
Is this issue related to this one #679  ?

elemoine:
> Is this issue related to this one #679 ?

Not really. But #697, which fixes this issue, removes `ngAnimate` entirely, which will fix #679!

PR #697, opened 2015-07-08
New "resize map" implementation

### LINKED ISSUE #679: ev.elapsedTime is undefined
When navigating on catalog I receive this error : (with ff, seems to work with chrome)

TypeError: ev.elapsedTime is undefined
http://devv3rm.geoportail.lu/node_modules/angular-animate/angular-animate.js
Line 1262

elemoine:
This PR hopefully fixes the map distortion issues once for good.

As [commented](https://github.com/Geoportail-Luxembourg/geoportailv3/issues/664#issuecomment-119567039) by @pgiraud, there are bugs in `ngAnimate` that prevent us from relying on the `$animate` service to know when animations start and end.

So we decided to write a simple directive that just, when a state changes, resizes the map in animation loop during 1 second. This may sound hackish but it is simple and more robust than other solutions. 

One benefit is the removal of `angular-animate`.

Fixes https://github.com/Geoportail-Luxembourg/geoportailv3/issues/664.
Fixes #679.


geoportailv3/static/js/locationinfo/locationinfodirective.js
```diff
@@ -179,24 +178,12 @@ app.LocationinfoController =
...
-   * @type {angular.Scope}
-   * @private
-   */
-  this.scope_ = $scope;
-
-  /**
    * @type {angular.$timeout}
    * @private
    */
   this.$timeout_ = $timeout;
```
pgiraud: Once the pull request is merged, I think we can remove this private variable.

pgiraud:
LGTM. Please merge.

PR #700, opened 2015-07-09
change cursor and perform query only one the previous one has finished

rmichaelis:
It can be tested on devv3rm


rmichaelis:
@petzlux could you please review ?

petzlux:
yes doing right now

PR #707, opened 2015-07-10
Translations - update of tools and plural translations

petzlux:
The angular gettext translation string extractor simply does string matching for finding translation strings.
This means that renamed methods such as `this.translate_` wont work for string extraction. 
````
this.translate_ = gettextCatalog;

gettextCatalog.getString('string1'); // gets extracted
gettext('string2'); // gets extracted
gettextCatalog.getPlural(n, 'singular', 'plural'); // gets extracted correctly as one single/plural pair string

var string3 = 'string3'; // string does not get extracted
gettextCatalog.getString(string3) // will not be translated as string is not extracted
this.translate_.getString(string3) // will not be translated

var string4 = gettext('string4'); // string gets extracted
gettextCatalog.getString(string4) // will be translated as string is extracted 
this.translate_.getString(string4) // will be translated
````

### DIFF
--- a/geoportailv3/locale/de/LC_MESSAGES/geoportailv3-client.po
+++ b/geoportailv3/locale/de/LC_MESSAGES/geoportailv3-client.po
@@ -1,11 +1,12 @@
 # 
 # Translators:
 # Andre aDonno <andrea.donno@me.com>, 2015
+# Francis Kaell <francis.kaell@act.etat.lu>, 2015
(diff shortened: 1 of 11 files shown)


petzlux:
Ready for review !

rmichaelis:
looks good to me.

petzlux:
LGTM

PR #723, opened 2015-07-14
DMS Coordinates output - add one decimal

### LINKED ISSUE #710: Coordinates DMS add one decimal

petzlux:
fix #710

### DIFF
--- a/geoportailv3/static/js/locationinfo/locationinfo.html
+++ b/geoportailv3/static/js/locationinfo/locationinfo.html
@@ -1,6 +1,6 @@
 <div class="container-fluid" ng-show="ctrl.appSelector == 'locationinfo'">
   <div class="row">
-    <div class="col-xs-8">
+    <div class="col-xs-8" style="padding-left: 0px;">
       <h3 translate>Short Url</h3>
       <input onclick="this.select();" type="text" class="form-control" name="shorturl" ng-model="ctrl.url" readonly="readonly" />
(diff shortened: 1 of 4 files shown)


petzlux:
review please

geoportailv3/static/js/locationinfo/locationinfo.html
```diff
@@ -1,6 +1,6 @@
 <div class="container-fluid" ng-show="ctrl.appSelector == 'locationinfo'">
   <div class="row">
-    <div class="col-xs-8">
+    <div class="col-xs-8" style="padding-left: 0px;">
```
rmichaelis: You should maybe use a class instead of a style

rmichaelis:
I think this is no more correct : 
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/petzlux_dms/geoportailv3/static/js/coordinatestringservice.js#L53

You have to replace ```var hdms = ol.coordinate.toStringHDMS(coordinate);``` with something like that ```var hdms = toStringHDMS_(coordinate);```

If not the decimal will never be displayed.

petzlux:
Ups ! I did a rebase and probably messed that change up there ...

geoportailv3/static/js/locationinfo/locationinfo.html
petzlux: okay!

PR #737, opened 2015-07-15
Fix scalebar length

sbrunner:


sbrunner:
Ready :-)

jaykayone:
I am testing it currently, give me 5 :)

jaykayone:
works great, please merge, thanks!

rmichaelis:
Why is the shortner url only working with POST and no more with GET ? Is there a reason ?

sbrunner:
Yes, the shortener will modify the database than It's recommend to do a `POST`, but it's true that not relay good to get the URL from the query string it will be cleaner id we get it from e JSON :-)

PR #739, opened 2015-07-16
Same layout as for the measure tools

rmichaelis:


rmichaelis:
@jaykayone Do we need a blank line or just the same layout as the one in measure tools (a transparent line) ?

jaykayone:
same layout

PR #740, opened 2015-07-16
clickable title and delete button

rmichaelis:
could you please check on mobile devices ?

### DIFF
--- a/geoportailv3/static/js/layermanager/layermanager.html
+++ b/geoportailv3/static/js/layermanager/layermanager.html
@@ -3,15 +3,13 @@
   <li ng-repeat="layer in ctrl.layers" class="panel layermanager-panel">
   <span class="sortable-handle fa fa-reorder"></span>
     <app-layerinfo app-layerinfo-layer="::layer"></app-layerinfo>
+    <div data-toggle="collapse" data-parent="#layermanager-{{::ctrl.uid}}" data-target="#layermanager-item-{{$index}}" class="layer-label">
     {{layer.get('label') | translate}}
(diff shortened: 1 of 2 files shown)


rmichaelis:
could be tested here : devv3.geoportail.lu/renaud

petzlux:
screenshot_2015-07-16-11-33-20

petzlux:
:-1:

rmichaelis:
Does it work with a layer having a shorter name ?

petzlux:
Nope

screenshot_2015-07-16-11-56-05

rmichaelis:
what browser are you using ?

petzlux:
chrome on android 5

jaykayone:
on iOS it works great

rmichaelis:
@petzlux is it better now ? It works fine on my android devices.

petzlux:
Yes works now!

rmichaelis:
As discussed I've just added the ellipsis right to the title.
Is this now ready to merge ?

petzlux:
I am of the opinion that we should hide the ellipsis on desktop , as this is just a hint for mobile / tablet users, that they can expand the layer tile ... this would simply be a css rule

jaykayone:
the dots are now too close to the bin and the text is not shown entirely



we should rediscuss this tomorrow

rmichaelis:
Do you have a screenshot ?

jaykayone:
now it is OK.
as far as I know, we currently do not have another option to detect mobile as the size, which does not really detect tablets correctly. 
In that regard I suggest that we should show the ellipses everywhere until we have a possbility to sniff the touch capability of a device

PR #758, opened 2015-07-22
Minor additions

petzlux:


jaykayone:
the result LGTM

the new update-translations settings makes uncommited changes everytime though so that I cannot easily switch branches

petzlux:
@jaykayone git reset --hard HEAD   then no problems switching branches. translations should anyway get committed and merged through pull requests like any other changes

jaykayone:
yes but they should not be automagically generated in build target

PR #762, opened 2015-07-23
changed behaviour of topbar theme shortcut

petzlux:


petzlux:
@jaykayone can you review and let me know if the behaviour is now as wished?

jaykayone:
OK 4 me

PR #771, opened 2015-08-05
Use git+https://github.com/ in package.json

elemoine:
With this commit we use https://github.com/xx/yy#master rather than xx/yy#master in package.json when specifying dependencies downloaded from GitHub. In this way, npm doesn't attempt to download the package from registry.npmjs.org; instead it goes directly to GitHub. This prevents a failure at npm install time when the content of the 404 response from registry.npmjs.org is empty. It also makes things a bit faster as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,15 +16,15 @@
     "jquery": "^1.9.1",
     "jshint": "2.7.0",
     "less": "~1.7.5",
-    "ngeo": "camptocamp/ngeo#master",
+    "ngeo": "git+https://github.com/camptocamp/ngeo#master",
     "nomnom": "~1.6.2",
-    "openlayers": "openlayers/ol3#master",
+    "openlayers": "git+https://github.com/openlayers/ol3#master",
     "proj4": "2.3.3",
     "phantomjs": "~1.9.7-5",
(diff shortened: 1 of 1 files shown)


petzlux:
@elemoine  this can cause problems with our proxy, as we need to use https not git download protocol for the downloads to pass.

Altough I can see that there is a global setting, will try this out tomorrow:

````git config --global url."https://".insteadOf git://````

petzlux:
@elemoine checked and indeed, with git: the install fails, given our restrictive proxy. 

But the git config mentioned did the trick !

elemoine:
We could possibly use `https` instead of `git`.

elemoine:
Now using `https` rather than `git`. I'll merge when Travis build passes.

elemoine:
It doesn't work on Travis, for some reason…

elemoine:
It seems to work when using git+https. @petzlux, does that work behind the proxy?

petzlux:
Build works , LGTM & LGTM ! (Looks Good To Me & Lets Get This Merged :-)

elemoine:
Thanks. I hope it'll work behind the proxy.

PR #816, opened 2015-08-11
Do not query the map while dragging the map

### LINKED ISSUE #769: Map performance issue on FF
I am observing major map pan performance issues in FF.

For example with http://devv3.geoportail.lu/main/wsgi/theme/main?version=3&zoom=16&X=680704&Y=6376939&lang=lb&layers=373-154-146-151&opacities=1-1-1-1&bgLayer=basemap_2015_global&debug

I just run Profiles test and it seems that `forEachLayerAtPixel` takes a lot of CPU time. So we could probably make map panning a lot faster by checking that the map is not being panned before calling `forEachLayerAtPixel`.

I am considering this a bug.

elemoine:
`map.forEachLayerAtPixel()` is a costly operation. We should not do that operation while the map is being dragged.

This makes panning the map much faster for me on FF Linux. And it should improve the pan performance in other browsers as well.

Please review.

Fixes #769.

### DIFF
--- a/geoportailv3/static/js/query/querydirective.js
+++ b/geoportailv3/static/js/query/querydirective.js
@@ -254,8 +254,8 @@ app.QueryController = function($sce, $timeout, $scope, $http,
 
   goog.events.listen(this.map_, ol.MapBrowserEvent.EventType.POINTERMOVE,
       function(evt) {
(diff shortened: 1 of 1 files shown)


rmichaelis:
Looks good to me

PR #818, opened 2015-08-11
Missing tooltips

rmichaelis:


rmichaelis:
@petzlux bus stop is now ok as well as the language in the link to the shop

PR #819, opened 2015-08-11
always create the tooltips pot file

rmichaelis:
@jaykayone could you please review and tell me if it works as you expect ?


petzlux:
LGTM

PR #832, opened 2015-08-27
update of translations

petzlux:


jaykayone:
merge or remove ?

petzlux:
update! after the other branches are merged , would be good to update translations here

PR #833, opened 2015-08-27
mymaps webserbices

rmichaelis:
Here is the first draft of the mymaps' webservices
Could you please review and tell me if things are missing or need to be adapted for the future integration. 

Thanks.


geoportailv3/__init__.py
```diff
@@ -44,6 +45,59 @@ def main(global_config, **settings):
         "/printproxy/cancel/{ref}",
         request_method="DELETE"
     )
+    # mymaps routes
+    config.add_route(
+        "mymaps_getmaps",
+        "/mymaps/getmaps"
```
elemoine: Why not use `/mymaps/maps`?

elemoine: And you could use `request_method="GET"` (here and elsewhere).

elemoine:
Do we execute the Python tests on Travis? It would be very regrettable if we did not :)

geoportailv3/mymaps.py
```diff
@@ -0,0 +1,321 @@
...
+    def __iter__(self):
+        """Returns an iterable that supports .next()
+            so we can do dict(sa_instance)
+        """
+        return self.todict()
+
+    @staticmethod
+    def get(id):
+        """ Get map by its id. """
+        return DBSession.query(Map).get(id)
```
elemoine: For consistency reasons I'd recommend directly using `DBSession.query(Map).get(id)` in the view code? Defining and using such static methods is not a common SQLAlchemy pattern.

elemoine: I see that you do the same in `belonging_to`. So let's forget about this. This is minor anyway.

geoportailv3/mymaps.py
```diff
@@ -0,0 +1,321 @@
...
+        self.stroke = feature.properties.get('stroke')
+        self.is_label = feature.properties.get('isLabel')
+        self.show_orientation = feature.properties.get('showOrientation')
+        self.linestyle = feature.properties.get('linestyle')
+
+        if (feature.properties.get('size') is not None and
+                unicode(feature.properties.get('size')).isnumeric()):
+            self.size = feature.properties.get('size')
+        else:
+            self.size = 40
```
elemoine: Here you repeat `feature.properties.get('size')` three times. How about the following:

```py
        size = feature.properties.get('size')
        if size is not None and unicode(size).isnumeric():
            self.size = size
        else:
            self.size = 40
```

or

```py
        size = feature.properties.get('size')
        self.size = size if size is not None and unicode(size).isnumeric() else 40
```

?

geoportailv3/mymaps.py
```diff
@@ -0,0 +1,321 @@
...
+            self.angle = 0
+
+        if (feature.properties.get('font_size') is not None and
+                unicode(feature.properties.get('font_size')).isnumeric()):
+            self.font_size = feature.properties.get('font_size')
+        else:
+            self.font_size = 15
+
+        if (feature.properties.get('symbol_id') is not None and
+                len(str(feature.properties.get('symbol_id'))) == 0):
```
elemoine: Why do you use `str` here and `unicode` for other properties?

geoportailv3/mymaps.py
```diff
@@ -0,0 +1,321 @@
...
+                    ob = geom
+                geom_type = ob.get("type").lower()
+
+                if types is None:
+                    types = geom_type
+                else:
+                    if types == geom_type:
+                        is_transformable = True
+                    else:
+                        is_transformable = False
```
elemoine:
```py
is_transformable = types == geom_type
```

?

geoportailv3/mymaps.py
```diff
@@ -0,0 +1,321 @@
...
+                if types == "point":
+                    shape = asMultiPoint(geoms)
+                elif types == "linestring":
+                    shape = asMultiLineString(geoms)
+                elif types == "polygon":
+                    shape = MultiPolygonAdapter(geoms, context_type='geojson')
+            else:
+                shape = None
+
+        self.geometry = from_shape(shape, srid=2169)
```
elemoine: `from_shape` throws a `NameError` when the shape is `None`.

geoportailv3/mymaps.py
```diff
@@ -0,0 +1,321 @@
...
+                          description=self.description,
+                          thumbnail=self.thumbnail,
+                          image=self.image,
+                          color=self.color,
+                          stroke=self.stroke,
+                          isLabel=self.is_label,
+                          showOrientation=self.show_orientation,
+                          linestyle=self.linestyle,
+                          id=self.id,
+                          symbol_id=self.symbol_id,
```
elemoine: For consistency reasons should we use `symbolId` here?

geoportailv3/mymaps.py
```diff
@@ -0,0 +1,321 @@
...
+
+    id = Column(Integer, primary_key=True)
+    name = Column(Unicode(255))
+    allow_labeling = Column(Boolean)
+
+    def __init__(self, name):
+        self.name = name
+
+    def __allow_labeling__(self):
+        if (self.allow_labeling):
```
elemoine: You can remove the brackets here.

geoportailv3/mymaps.py
```diff
@@ -0,0 +1,321 @@
...
+            return True
+        else:
+            return False
```
elemoine: Ad what's the point of that function? Why just not use `self.allow_labeling` in `todict`?

geoportailv3/views/mymaps.py
```diff
@@ -0,0 +1,622 @@
...
+                END
+            """ % (symbol.id, str(symbol.id) + ext)
+
+        the_file = open(dir+"/symbols.map", 'w+')
+        the_file.write(symbolsmap)
+        the_file.close()
+        os.system('scp "%s" "%s:%s"'
+                  % (dir + "/symbols.map",
+                     "mapserver@todd.geoportal.lu",
+                     "/home/mapserver/config/MYMAPS/symbols"))
```
elemoine: It was done this way before?

elemoine:
Done reviewing the code for now.

PR #835, opened 2015-08-27
fix ngeo bloodhound service after update

petzlux:


elemoine:
Sorry about that. I should have notified you about this change.

PR #836, opened 2015-08-31
Lux muliselection

rmichaelis:
@jaykayone Is it working as expected ? 
Maybe could we use the same mechanism to order one cadastral extract for multiple parcels.
Please review.

### DIFF
--- a/geoportailv3/static/js/query/querydirective.js
+++ b/geoportailv3/static/js/query/querydirective.js
@@ -56,6 +56,12 @@ app.QueryController = function($sce, $timeout, $scope, $http,
     getRemoteTemplateServiceUrl) {
 
   /**
+   * @type {Array}
+   * @private
+   */
+  this.responses_ = [];
+
+  /**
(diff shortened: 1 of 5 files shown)


geoportailv3/static/js/query/querydirective.js
```diff
@@ -398,6 +438,23 @@ app.QueryController.prototype.hasAttributes = function(feature) {
...
+ * @param {string} sep the join separator
+ * @return {string} the string with joined attributes
+ * @export
+ */
+app.QueryController.prototype.joinAttributes = function(features, attr, sep) {
+  var attributes = [];
+  for (var i = 0; i < features.length; i++) {
+    attributes.push(features[i].attributes[attr]);
+  }
+  return attributes.join(sep);
```
elemoine: You could use

```js
  return goog.array.map(features, function(feature) {
    return feature.attributes[attr];
  }).join(sep);
```

PR #842, opened 2015-09-01
Use c2c.template>=1.2.dev in CONST_dev-requirements.txt

elemoine:
This fixes a wheel-related problem that occurs when using recent versions of pip. Recent pip versions indeed refuse to install `zc.recipe.egg`, with this error: `zc.recipe.egg is in an unsupported or invalid wheel`. See https://github.com/camptocamp/c2cgeoportal/issues/1494.

This change anticipates on changes that will be applied when updating the project to a more recent version of c2cgeoportal. See https://github.com/camptocamp/c2cgeoportal/commit/dc33e73de381ec16fd51471ac7edbd37d526824b.


elemoine:
I also converted the `z3c.recipe.filetemplate` templates (`.in`) into Mako templates (`.mako`). In this way we don't depend on `z3c.recipe.filetemplate`, so we don't pull `zc.recipe.egg` and `zc.buildtout`. It is also to be noted that this is consistent with recent changes to c2cgeoportal.

pgiraud:
Thanks for the explanations. This looks good to me. Please merge.

PR #844, opened 2015-09-02
Use closure-compiler#master for externs again

elemoine:
OpenLayers master now uses the most recent version of Closure Compiler. See https://github.com/openlayers/ol3/pull/4072. So we can use the most recent version of the Angular and jQuery externs again.


elemoine:
No functional change, and fixes build problems. Merging.

PR #846, opened 2015-09-02
Remove one-time binding in catalog

### LINKED ISSUE #837: Catalog is not updated
When switching from one theme to another, the catalog is not updated.

elemoine:
This fixes a regression introduced by d42e5a4, where the catalog tree is not updated when switching themes. This commit removes the use of one-time binding that causes that problem.

Fixes #837.


rmichaelis:
works fine. Looks good to me.

PR #848, opened 2015-09-02
Use template cache also in debug mode

### LINKED ISSUE #826: Theme switcher collapse/expand doesn't work correctly in debug mode
Strangely enough, the collapsing/expanding of the theme switcher does not correctly work in debug mode (with `&debug` in the URL).

This is easily reproducible on http://devv3.geoportail.lu/main/wsgi/?debug. Just open the theme switcher and this is what you'll get:

themeswitcher-problem

elemoine:
See commit message for details.

Fixes #826.


elemoine:
This depends on https://github.com/Geoportail-Luxembourg/geoportailv3/pull/842, which should be merged first.

pgiraud:
Will we have to watch for modifications on templates like we do for less files?

elemoine:
> Will we have to watch for modifications on templates like we do for less files?

Yes indeed.

A change to one of the directive partials requires running `make geoportailv3/static/build/templatecache.js`.

pgiraud:
OK. Looks good to me. Please merge.

PR #854, opened 2015-09-03
Ignore the generated *-client.po files

pgiraud:
I just built the project and got "not staged changes" ready to commit on ".po" files.
I think they need to be ignored.
Please review.


elemoine:
I always have the same problem, but I honestly don't know what needs to be done.

elemoine:
I agree with merging this. And we'll see if @petzlux agrees when he comes back.

PR #863, opened 2015-09-10
migrate to typehead 0.11.1

rmichaelis:
Could you please review.

My biggest problem was the fact that dataset is no more transmitted to the selected event. 
The way I've used to solved that problem is putting the dataset name into each result. 
But unfortunately, I can't do that here https://github.com/Geoportail-Luxembourg/geoportailv3/blob/fix_search/geoportailv3/static/js/search/searchdirective.js#L383

Probably not perfect but it seems to work. If you have any clues to improve that, you are welcome.

FYI This is no more correct : 
https://github.com/camptocamp/ngeo/blob/master/externs/ngeox.js#L315,L316

### DIFF
--- a/geoportailv3/static/js/search/searchdirective.js
+++ b/geoportailv3/static/js/search/searchdirective.js
@@ -28,7 +28,9 @@ goog.require('ngeo.searchDirective');
 
 
 /**
- * @typedef {{bgLayer: ol.layer.Layer, translatedName: string}}
+ * @typedef {{
+ *  dataset: string,
+ *  result: {bgLayer: ol.layer.Layer, translatedName: string}}}
  */
 app.BackgroundLayerSuggestion;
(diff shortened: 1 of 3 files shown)


elemoine:
I came across the same problems as you and decided to go back to 0.10.5. And now you're moving back to 0.11.1 :)

elemoine:
As discussed, let's close this. We're going to stick to typeahead 0.10.5 for now, as typeahead 0.11.1 has too many problems/bugs for us. The thing we still need to do is implement our own `CreateGeoJSONBloodhound` factory.

elemoine:
@rmichaelis, let's keep your fix_search branch for now. We can delete it when we have a fully working search component.

elemoine:
I guess you can now delete the branch.

PR #866, opened 2015-09-16
Feature popup

pgiraud:
This pull request main goal is to display a popup when a feature is selected (on the map or via the list in mymaps block).
As for now, the popup shows name and title in a read only mode and user is able to modify those attributes.

Next steps are :
 - add style editing,
 - enable feature geometry modification,
 - allow user to remove feature.

A first review would be much appreciated.


rmichaelis:
The query information has to be deactivated during the drawing, if not each click will return information about the queryable features on the map.
See how it was done in measure :
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/js/measure/measuredirective.js#L281

rmichaelis:
I'm not sure, but I have the feeling that double-clicking on a map element in the left list, should allow to edit the name.
I have this feeling because clicking on the map's title or description opens the modification popup.
It's probably a bad idea.

rmichaelis:
If you are drawing an element, you should deactivate the feature selection on click

rmichaelis:
When you finish to draw an element, the element popup should be opened.

geoportailv3/templates/index.html
```diff
@@ -145,7 +145,7 @@ <h2 translate>my_maps</h2>
             ✕
           </button>
           <div class="col-xs-12">
-            <app-mymaps app-mymaps-selectinteraction="::mainCtrl.selectInteraction"
+            <app-mymaps app-mymaps-selectedfeatures="::mainCtrl.selectedFeatures"
               app-mymaps-features="::mainCtrl.drawnFeatures">
             <app-mymaps>
```
rmichaelis: Closing tag is missing

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -132,6 +168,189 @@ app.DrawController = function($scope, ngeoDecorateInteraction) {
...
   }, this));
+
+
+  /**
+   * @param {number} resolution The map resolution.
+   * @this {ol.Feature}
+   * @return {Array<ol.style.Style>}
+   * @private
+   */
+  this.styleFunction_ = function(resolution) {
```
elemoine: Replace with:

```js
this.styleFunction_ = app.DrawController.createStyleFunction_();
```
And then the `createStyleFunction_` would look like this:

```js
app.DrawController.createStyleFunction_ = function() {
  var styles = [];
  // …
  return function(resolution) {
    styles.length = 0;
    styles.push(...);
    return styles;
  };
};
```

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -132,6 +168,189 @@ app.DrawController = function($scope, ngeoDecorateInteraction) {
...
+            coordinates = feature.getGeometry().getCoordinates();
+            return new ol.geom.MultiPoint(coordinates);
+          } else if (geom instanceof ol.geom.Polygon) {
+            coordinates = feature.getGeometry().getCoordinates()[0];
+            return new ol.geom.MultiPoint(coordinates);
+          } else {
+            return feature.getGeometry();
+          }
+        }
+      }));
```
elemoine: This style object could be created outside the style function.

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -132,6 +168,189 @@ app.DrawController = function($scope, ngeoDecorateInteraction) {
...
+
+    // goog.asserts.assert(goog.isDef(this.get('__style__'));
+    var style = this.get('__style__');
+    var color = style['color'];
+    var rgb = goog.color.hexToRgb(color);
+    var fillColor = goog.color.alpha.rgbaToRgbaStyle(rgb[0], rgb[1], rgb[2],
+        style['opacity']);
+    var fill = new ol.style.Fill({
+      color: fillColor
+    });
```
elemoine: The `fill` object could be created once for good outside the style function.

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -132,6 +168,189 @@ app.DrawController = function($scope, ngeoDecorateInteraction) {
...
+      });
+
+  goog.events.listen(this.selectInteraction,
+      ol.interaction.SelectEventType.SELECT,
+      /**
+       * @param {ol.interaction.SelectEvent} evt
+       */
+      function(evt) {
+        if (evt.selected.length > 0) {
+          this.featurePopup_.show(evt.selected[0],
```
elemoine: Adding `var feature = evt.selected[0]` would make the code easier to read.

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -140,7 +359,19 @@ app.DrawController = function($scope, ngeoDecorateInteraction) {
...
+  feature.set('__editable__', true);
+  feature.set('__style__', {
+    color: '#ed1c24',
+    opacity: 0.2,
+    size: feature.getGeometry().getType() == ol.geom.GeometryType.POINT ?
+        10 : 1.25,
+    shape: 'circle',
+    rotation: 0,
+    lineDash: 'plain'
+  });
```
elemoine: See https://github.com/Geoportail-Luxembourg/geoportailv3/pull/833/files for the list of feature properties expected by the mymaps web service.

geoportailv3/static/js/draw/styleediting.html
```diff
@@ -0,0 +1,74 @@
...
+    <table class="palette">
+      <tr ng-repeat="colors in ctrl.colors">
+        <td ng-repeat="color in colors" style="background-color:{{color}}" ng-click="ctrl.setColor(color)" ng-class="{'selected': color == ctrl.color}">
+        </td>
+      </tr>
+    </table>
+  </div>
+</div>
+
+<div class="form-inline" ng-show="ctrl.type == 'polygon'">
```
elemoine: You could possibly use a one-time binding expression here.

geoportailv3/static/js/draw/styleediting.html
```diff
@@ -0,0 +1,74 @@
...
+<div class="form-inline" ng-show="ctrl.type == 'linestring'">
+  <!-- Stroke size lines -->
+  <div class="form-group">
+    <label>Stroke width</label>
+    <input type="range" min="1" max="10" class="form-control" ng-model="ctrl.getSetSize" ng-model-options="{getterSetter: true}">
+  </div>
+  <div class="form-group">
+    <!-- Line style -->
+    <label>Style</label>
+    <button ng-click="ctrl.setLineDash('plain')" ng-class="{'selected': ctrl.lineDash == 'plain'}">Plain</button>
```
elemoine: You don't need quotes around `selected` here.

geoportailv3/static/js/draw/styleediting.html
```diff
@@ -0,0 +1,74 @@
...
+    <button ng-click="ctrl.setLineDash('dashed')" ng-class="{'selected': ctrl.lineDash == 'dashed'}">Dashed</button>
+    <button ng-click="ctrl.setLineDash('dotted')" ng-class="{'selected': ctrl.lineDash == 'dotted'}">Dotted</button>
+  </div>
+</div>
+<div class="form-inline" ng-show="ctrl.type == 'point'">
+  <div class="form-group">
+    <!-- Symbol style -->
+    <label>Symbol</label>
```
elemoine: You need to use the `translate` directive here.

geoportailv3/static/js/draw/styleediting.js
```diff
@@ -0,0 +1,224 @@
...
+   * @export
+   */
+  this.feature;
+
+
+  /**
+   * @type {string}
+   * @export
+   */
+  this.type;
```
elemoine: If possible assign values to the object properties.

geoportailv3/static/js/draw/styleediting.js
```diff
@@ -0,0 +1,224 @@
...
+  }, this));
+};
+
+
+/**
+ * @param {string} lineDash
+ * @export
+ */
+app.StyleEditingController.prototype.setLineDash = function(lineDash) {
+  if (!goog.isDefAndNotNull(this.feature)) {
```
elemoine: You could use `goog.isDef` instead of `goog.isDefAndNotNull` if you made sure `this.feature` was never set to `null`.

geoportailv3/static/js/featurepopup/featurepopupservice.js
```diff
@@ -0,0 +1,115 @@
...
+  this.scope_['feature'] = feature;
+  var anchor = goog.isDef(opt_anchor) ? opt_anchor : this.getAnchor(feature);
+  this.overlay_.setPosition(anchor);
+};
+
+
+/**
+ */
+app.FeaturePopup.prototype.hide = function() {
+  this.scope_['feature'] = null;
```
elemoine: Set `this['feature']` to `undefined. In this way you can use `goog.isDef` instead of `goog.isDefAndNotNull`.

geoportailv3/static/js/draw/styleediting.html
pgiraud: I can't actually since the same directive is used for every feature.

pgiraud:
I changed the code to do the same in the drawing tools.
However, we think that it's a bad idea to do it this way. The tools
(measure or drawing) shouldn't be aware of any query tool. Instead they
should probably notify the main controller that they are active.

Also, there may be a usability issue. When the draw toolbar is open but no
tool is active, the user may want to query the map, but nothing will happen
until he closes the draw toolbar. Hard to find probably. To be tested.

On Thu, Sep 17, 2015 at 5:26 PM, Renaud Michaëlis <notifications@github.com>
wrote:

> The query information has to be deactivated during the drawing, if not
> each click will return information about the queryable features on the map.
> See how it was done in measure :
>
> https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/js/measure/measuredirective.js#L281
>
> —
> <https://github.com/Geoportail-Luxembourg/geoportailv3/pull/866#issuecomment-141121909>
> .
>



-- 
Pierre GIRAUD
Géomaticien, Analyste

Camptocamp France SAS
Savoie Technolac, BP 352

Tel : 00 33 4 79 44 44 93
Mail : pierre.giraud@camptocamp.com
http://www.camptocamp.com

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -82,9 +98,21 @@ app.DrawController = function($scope, ngeoDecorateInteraction, ngeoLocation) {
 
   /**
    * @type {ol.Collection.<ol.Feature>}
-   * @export
+   * @private
    */
-  this.features;
+  this.features_ = appDrawnFeatures;
```
elemoine: I'd suggest renaming this `this.drawFeatures_` for more clarity.

geoportailv3/static/js/draw/drawnfeaturesservice.js
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * @typedef {ol.Collection<ol.Feature>}
+ */
+app.DrawnFeatures;
+
+
+/**
+ * The selected features collection.
+ * @private
```
elemoine: Very minor: you can add `@type {app.DrawFeatures}`, and `@const` to tell the compiler that this won't change.

geoportailv3/static/js/draw/styleediting.js
```diff
@@ -0,0 +1,210 @@
...
+ * @export
+ */
+app.StyleEditingController.prototype.getSetStroke = function(val) {
+  if (!goog.isDef(this.feature)) {
+    return;
+  }
+  if (arguments.length) {
+    this.feature.set('stroke', parseFloat(val));
+    this.feature.changed();
+    return;
```
elemoine: Can't you remove this `return` statement?

geoportailv3/static/js/featurepopup/featurepopupservice.js
```diff
@@ -0,0 +1,115 @@
...
+/**
+ * @param {ol.Feature} feature
+ * @return {?ol.Coordinate}
+ */
+app.FeaturePopup.prototype.getAnchor = function(feature) {
+  var geometry = feature.getGeometry();
+  switch (geometry.getType()) {
+    case ol.geom.GeometryType.POINT:
+      goog.asserts.assertInstanceof(geometry, ol.geom.Point,
+          'geometry should be an ol.geom.Point');
```
elemoine: I wouldn't add messages here, as the assertion is here basically just to please the compiler.

geoportailv3/static/js/maincontroller.js
```diff
@@ -180,12 +182,11 @@ app.MainController = function(
...
   /**
-   * The list of features that were drawn (either using the drawing tools or by
-   * loading a mymap map).
-   * @type {ol.Collection.<ol.Feature>}
+   * The selected features.
+   * @type {ol.Collection.<ol.Feature>?}
    * @export
    */
-  this.drawnFeatures = new ol.Collection();
+  this.selectedFeatures;
```
elemoine: To be removed, now there's a service.

elemoine:
I've added a couple more comments, but it looks very good to me. Please merge when you think this is ready.

geoportailv3/templates/index.html
```diff
@@ -312,6 +311,7 @@ <h2 translate>infos</h2>
          appModule.value('appQueryTemplateUrl', 'query/info.html');
          appModule.value('appLocationinfoTemplateUrl', 'locationinfo/locationinfo.html');
          appModule.value('appMymapsTemplateUrl', 'mymaps/mymaps.html');
+         appModule.value('appFeaturePopupTemplateUrl', 'featurepopup/featurepopup.html');
```
elemoine: This is the mistake.

Issue #845, opened 2015-09-02
permalink vs mymaps id


jaykayone:
if a user draws something outside a mymaps map, the features are stored in the permalink

if a user draws something in a map he has write rights on, the features are stored in the map and the map_id goes into the permalink

if a user draws something in a map he has no write rights on, the features are stored in the permalink and the map_id goes into the permalink too

rmichaelis:
waiting for #882

PR #870, opened 2015-09-21
Permalink feature encoding/decoding

elemoine:
This PR changes the draw directive to encode drawn features in the URL. At init time the directive also decodes the features set in the URL and adds them to the collection of drawn features.

Two things to note:

* At this point features are not added to the local storage. I don't think this is something we've discussed yet.
* #845 is not implemented yet, as there is no notion of mymaps map in the code at this point.


elemoine:
@rmichaelis, could you please review this? Thanks.

rmichaelis:
It looks really good to me.

PR #871, opened 2015-09-22
Use our own CreateGeoJSONBloodhound factory

elemoine:
This PR should fix the remaining search problems we have.

@rmichaelis, can you please test this?

### DIFF
--- a/build.json
+++ b/build.json
@@ -20,7 +20,6 @@
       "node_modules/ngeo/externs/d3.js",
       "node_modules/ngeo/externs/mapfish-print-v3.js",
       "node_modules/ngeo/externs/twbootstrap.js",
-      "node_modules/ngeo/externs/typeahead.js",
       "node_modules/ngeo/externs/closure-compiler.js",
       ".build/externs/angular-1.4.js",
       ".build/externs/angular-1.4-q_templated.js",
--- a/geoportailv3/externs/typeahead.js
+++ b/geoportailv3/externs/typeahead.js
@@ -1,12 +1,172 @@
 /**
- * @fileoverview Extern definitions for typeahead.
+ * @fileoverview Externs for Typeahead and Bloodhound 0.10.5
+ * @see https://github.com/twitter/typeahead.js/blob/master/doc/jquery_typeahead.md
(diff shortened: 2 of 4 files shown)


rmichaelis:
Results are now returned but the centering on the selected feature is not working properly.
image

elemoine:
> Results are now returned but the centering on the selected feature is not working properly.

Thanks for testing. What's the problem exactly?

rmichaelis:
I my example the feature should be displayed here 6.112554342579504 , 49.6106117525195
but it's displayed here 0.00006 0.00049

elemoine:
Strange. Do you have the same problem in debug mode?

rmichaelis:
Yes I have the same problem in debug mode. 
I've just executed a clean but the problem is still remaining.

elemoine:
Ok thanks.

Could you please try this patch?

```diff
diff --git a/geoportailv3/static/js/search/searchdirective.js b/geoportailv3/static/js/search/searchdirective.js
index e5d680d..d3b9726 100644
--- a/geoportailv3/static/js/search/searchdirective.js
+++ b/geoportailv3/static/js/search/searchdirective.js
@@ -529,7 +529,7 @@ app.SearchDirectiveController.prototype.createAndInitPOIBloodhound_ =
     function(appCreateGeoJSONBloodhound, searchServiceUrl) {
   var url = searchServiceUrl + '?limit=5&query=%QUERY';
   var bloodhound = appCreateGeoJSONBloodhound(
-      url, undefined, ol.proj.get('EPSG:3857'));
+      url, ol.proj.get('EPSG:3857'));
   bloodhound.initialize();
   return bloodhound;
 };
```

(The problem may be related to an old bug.)

elemoine:
The above patch assumes that the coordinates received in the GeoJSON document from the search webservice are longitudes/latitudes. Can you please confirm that this is the case?

rmichaelis:
With the patch it works as expected.
Thanks.

elemoine:
I'll add a commit and merge. Thanks!

elemoine:
New commit added. I'll merge when the build passes on Travis.

PR #877, opened 2015-09-24
Do not use checkStructDictInheritance

elemoine:
This compile option does not exist anymore. And we get a compile error when using a recent version of the compiler.


elemoine:
Note that the build will fail without this PR. So it should be merged sooner rather than later :)

PR #895, opened 2015-10-20
Set TEMPLATE_EXCLUDE in elemoine.mk

elemoine:
This is to prevent CONST_Makefile from searching the "tomcat" directory.


elemoine:
elemoine-only. Merging.

PR #898, opened 2015-10-20
Prevent property renaming in print directive

elemoine:
The `layout` and `scale` properties of `app.PrintController` are defined using the bracket notation. For example, this is how the `layout` property is defined in the constructor:

```js
/**
 * @type {string}
 */
this['layout'] = this['layouts'][0];
```

So when accessing these properties the bracket notation must be used as well. Do not ever mix the bracket and dot notations for a given property.


elemoine:
@rmichaelis, I'll merge this if you understand the issue and agree with the correction. Thanks.

rmichaelis:
Yes of course. I understand and I agree with your changes. 
Please merge.

PR #912, opened 2015-10-26
Petzlux create map

### LINKED ISSUE #888: usability
I fear that nobody will find how to edit the title and description of a map.

@pgiraud 

http://devv3.geoportail.lu/renaud/theme/main?version=3&zoom=13&X=686645&Y=6397626&lang=lb&layers=305&opacities=1&bgLayer=basemap_2015_global&map_id=864a60f9ad7c48799f5a6a0d3117397f&crosshair=false

petzlux:
@elemoine This code should work, but when I open the Modal, i get an error: 

````
TypeError: shownSet is not a function
    at modal.js:66
    at Scope.$eval (angular.js:15830)
    at Scope.$apply (angular.js:15929)
    at HTMLDivElement.<anonymous> (modal.js:65)
    at HTMLDivElement.jQuery.event.dispatch (jquery.js:4670)
    at HTMLDivElement.elemData.handle (jquery.js:4338)
    at Object.jQuery.event.trigger (jquery.js:4579)
    at HTMLDivElement.<anonymous> (jquery.js:5289)
    at Function.jQuery.extend.each (jquery.js:384)
    at jQuery.fn.jQuery.each (jquery.js:136)
````

Is this connected to how the `ngeo-modal-shown` angular expression is evaluated? 

fix #888

### DIFF
--- a/geoportailv3/static/js/mymaps/mymaps.html
+++ b/geoportailv3/static/js/mymaps/mymaps.html
@@ -102,6 +102,12 @@
     <div class="form-group">
       <label translate>Titre de la carte</label>
       <input type="text" class="form-control" ng-model="ctrl.newTitle">
+      <div class="checkbox">
+        <label>
+          <input type="checkbox" ng-model="ctrl.newIsPublic">
+          <span translate>Make Map Public</span>
+        </label>
+      </div>
       <div class="dropdown">
         <button type="button" class="btn btn-default" data-toggle="dropdown">
           {{ ctrl.getMapCategory(ctrl.newCategoryId).name }}<span class="caret"></span>
--- a/geoportailv3/static/js/mymaps/mymaps.html
+++ b/geoportailv3/static/js/mymaps/mymaps.html
@@ -3,10 +3,10 @@
     Create, save and share your own maps.
   </p>
   <p class="text-center">
-    <button class="btn btn-default" ng-click="ctrl.chooseMap()" translate>Ouvrir une carte</button>
+    <button class="btn btn-default" ng-click="ctrl.openChooseMapModal()" translate>Ouvrir une carte</button>
   </p>
   <p class="text-center">
-    <button class="btn btn-default" ng-click="ctrl.createMap()" translate>Créer une nouvelle carte</button>
+    <button class="btn btn-default" ng-click="ctrl.openCreateMapModal()" translate>Créer une nouvelle carte</button>
   </p>
 </div>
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: geoportailv3/static/js/mymaps/mymaps.html [hunk 1]
```diff
@@ -50,43 +50,82 @@
 </div>
 
 <!-- Modification Modal -->
-<ngeo-modal ngeo-modal-shown="ctrl.modifying">
+<ngeo-modal ngeo-modal-shown="ctrl.modifying || ctrl.creating">
```
petzlux: See here for where the problem lies

elemoine: The expression passed to `ngeo-modal-shown` must be assignable. `ctrl.modifying` is assignable while `ctrl.modifying || ctrl.creating` is not.

The "assignable" property is important because the `ngeo-modal` directive changes the value of the "shown" property when the user closes the modal.

elemoine: In your case you don't need to know when the user closes the modal?

petzlux: Okay! I will create a third variable then. Thanks

petzlux: As I am reusing the modifying template, I need to know if the modal is creating a new map or modifying an existing one ...

elemoine: Really, I think we should change the ngeo-modal to use ng-model.

petzlux: Sorry being dense again, can you please explain?

elemoine: You should be able to use `ngeo-modal` like this:

```html
<ngeo-modal ng-model="ctrl.modalShownHidden" ng-model-options="{getterSetter: true}">
  // …
</ngeo-modal>
```
And:

```js
/**
 * @param {boolean|undefined}
 * @return {boolean|undefined}
 * @export
 */
app.MymapsDirectiveController.prototype.modalShownHidden = function(value) {
  if (goog.isDef(value)) {
    // …
  } else {
    return this.modifying || this.changing;
  }
};
```

petzlux: okay thanks ! that made it clear :-)

elemoine: See https://github.com/camptocamp/ngeo/pull/372.

petzlux: okay , even clearer now ! waiting for the merge then in ngeo

elemoine: Solving a problem by adding state variables never is a good sign. Single source of trust for the win!

petzlux: So as I understand, I should create a enum: 
````
/**
* @enum {number}
*/
app.modalType = {
  HIDDEN: 0,
  CREATE: 1,
  MODIFY: 
};
...
/**
* @type {app.modalType}
* @export
**/
this.modal = app.modalType.HIDDEN;
````
Which I can then set `this.modal = app.modalType.CREATE`

Comment (elemoine) [inline geoportailv3/static/js/mymaps/mymaps.html]: I don't understand why you want to do that, but I may not have the full context.

To me the question is: do you need to define a specific action when the user closes the modal? For example, do you need to reset `modifying` or `creating` (or both) to `false`?

**Comment by elemoine** [comment]
For your information, https://github.com/camptocamp/ngeo/pull/372 is merged. Meaning that the modal markup needs to be changed in the project.

```html
<ngeo-modal ngeo-modal-shown="ctrl.modalShown">
  <!-- modal content -->
</ngeo-modal>
```

to be changed to:

```html
<ngeo-modal ng-model="ctrl.modalShown">
  <!-- modal content -->
</ngeo-modal>
```

geoportailv3/static/js/mymaps/mymaps.html
petzlux: @elemoine how do you propose to reuse the same modal form code to both create a new map, as well as modify an existing one. the information if the user wants to create a new map or modify an existing one has to be expressed somehow?

petzlux:
okay another try, please review ...

PR #920, opened 2015-10-28
Pgiraud mymaps database config

pgiraud:


pgiraud:
User only modification. Merging.

PR #929, opened 2015-10-30
Use htmlmin in templatecache.mako.js

elemoine:
This PR changes `templatecache.mako.js` to use the `htmlmin` package rather than minifying the HTML code ourselves.

The original problem was found by @pgiraud, where our code just blindly removed newline characters.

And

```html
<button>test</button>
<button>test</button>
```

is not the same as

```html
<button>test</button><button>test</button>
```

@pgiraud, could you please test this?

### DIFF
--- a/CONST_Makefile
+++ b/CONST_Makefile
@@ -515,7 +515,7 @@ $(OUTPUT_DIR)/build.css: $(LESS_FILES) .build/node_modules.timestamp
 		.build/node_modules.timestamp
 	./node_modules/openlayers/node_modules/.bin/closure-util build $< $@
 
-$(OUTPUT_DIR)/templatecache.js: templatecache.mako.js $(APP_PARTIALS_FILES)
+$(OUTPUT_DIR)/templatecache.js: templatecache.mako.js $(APP_PARTIALS_FILES) .build/dev-requirements.timestamp
 	mkdir -p $(dir $@)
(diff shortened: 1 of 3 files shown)


CONST_Makefile
```diff
@@ -515,7 +515,7 @@ $(OUTPUT_DIR)/build.css: $(LESS_FILES) .build/node_modules.timestamp
 		.build/node_modules.timestamp
 	./node_modules/openlayers/node_modules/.bin/closure-util build $< $@
 
-$(OUTPUT_DIR)/templatecache.js: templatecache.mako.js $(APP_PARTIALS_FILES)
+$(OUTPUT_DIR)/templatecache.js: templatecache.mako.js $(APP_PARTIALS_FILES) .build/dev-requirements.timestamp
```
sbrunner: Why adding this dependency?

elemoine: Because it includes the `htmlmin` dependency definition. And `htmlmin` should be installed prior to generating the template cache.

sbrunner: so basic, I miss this ..., thanks @elemoine :-)

pgiraud:
I can confirm that it works. Space between buttons is now correct. Thanks.

pgiraud:
Needs to be rebased.

elemoine:
Now rebased. I'll merge if the build passes on Travis.

PR #931, opened 2015-11-02
Remove config vars to avoid errors with c2c-template

pgiraud:
With a recent version of c2c-template being more strict on variables, the build doesn't pass.

The two missing config variables are "layers" and "reset_password".
I don't know if removing them from `CONFIG_VARS` in `CONST_Makefile` is a good idea, though.
@sbrunner can you please confirm?

Note: this is a blocker bug for now.

### DIFF
--- a/CONST_vars.yaml
+++ b/CONST_vars.yaml
@@ -71,6 +71,9 @@ vars:
     closure_library_path: 'process.stdout.write(require("openlayers/node_modules/closure-util").getLibraryPath())'
     node_modules_path: "{directory}/node_modules"
 
+    layers: {}
+    reset_password: {}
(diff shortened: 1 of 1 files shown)


sbrunner:
A good solution is to add them to `CONST_vars.yaml` because in some cases we needs them!

pgiraud:
What values should those variables take?

sbrunner:
Empty dictionary (`{}`) :-)

pgiraud:
@sbrunner Can you please confirm that is OK now? I would like to merge this one as soon as possible.

sbrunner:
Looks good :-)

PR #936, opened 2015-11-13
In case of a null value it returns a default value

### LINKED ISSUE #887: there are a lot of features that are not visible
probably because of an empty size field (when coming from mobile APP).
We should have a default fallback size

rmichaelis:
fixes #887


rmichaelis:
Please review

PR #954, opened 2015-12-01
Mymaps improved element naming

### LINKED ISSUE #921: Give the drawn features a relevant title
Currently, the drawn features are automatically given "element xx" as title.
It would be better in terms of usability to give them a more concrete name such as "Line 1" or "Point xx".

Also the counter should take the already drawn features in case of a loaded map. Currently, the counter starts from "0" even if features are already in the map. It may results in two features having the same title ( "element 1" for example).

petzlux:
fix #921

### DIFF
--- a/geoportailv3/static/js/draw/drawdirective.js
+++ b/geoportailv3/static/js/draw/drawdirective.js
@@ -69,13 +69,20 @@ app.module.directive('appDraw', app.drawDirective);
  * @param {app.DrawnFeatures} appDrawnFeatures Drawn features service.
  * @param {app.SelectedFeatures} appSelectedFeatures Selected features service.
  * @param {app.Mymaps} appMymaps Mymaps service.
+ * @param {angularGettext.Catalog} gettextCatalog Gettext service.
  * @constructor
  * @export
  * @ngInject
  */
 app.DrawController = function($scope, ngeoDecorateInteraction,
     ngeoFeatureOverlayMgr, appFeaturePopup, appDrawnFeatures,
-    appSelectedFeatures, appMymaps) {
+    appSelectedFeatures, appMymaps, gettextCatalog) {
+
+  /**
+   * @type {angularGettext.Catalog}
+   * @private
+   */
+  this.translate_ = gettextCatalog;
 
   /**
    * @type {ol.Map}
@@ -88,12 +95,6 @@ app.DrawController = function($scope, ngeoDecorateInteraction,
    * @export
    */
   this.active;
-
-  /**
-   * @type {number}
-   * @private
-   */
-  this.featureSeq_ = 0;
 
   /**
    * @type {app.DrawnFeatures}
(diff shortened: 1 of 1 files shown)


petzlux:
Review please

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -366,7 +373,33 @@ app.DrawController.prototype.onDrawEnd_ = function(event) {
...
+    case 'Polygon':
+      name = this.translate_.getString('Polygon');
+      break;
+  }
+  feature.set('name', name + ' ' + (++this.featureSeq_));
+  this.drawnFeatures_.features.forEach(function(drawnFeature) {
+    if (feature.get('name') === drawnFeature.get('name')) {
+      feature.set('name', name + ' ' + (++this.featureSeq_));
+    }
+  }, this);
```
pgiraud: What about initializing the sequence to the number of features + 1?
What I mean is that we could give the new feature a name with an index corresponding to the number of features already drawn.
```
feature.set('name', name + ' ' + this.drawnFeatures_.getCollection().getLength() + 1);
```
This is how google mymaps works.
Then, we can get rid of `featureSeq`.

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -366,7 +373,33 @@ app.DrawController.prototype.onDrawEnd_ = function(event) {
...
+      break;
+    case 'LineString':
+      name = this.translate_.getString('LineString');
+      break;
+    case 'Polygon':
+      name = this.translate_.getString('Polygon');
+      break;
+  }
+  feature.set('name', name + ' ' + (++this.featureSeq_));
+  this.drawnFeatures_.features.forEach(function(drawnFeature) {
```
pgiraud: The `features` should be private in `drawnfeaturesservice` and we should use `getCollection` instead.

geoportailv3/static/js/draw/drawdirective.js
```diff
@@ -366,7 +373,33 @@ app.DrawController.prototype.onDrawEnd_ = function(event) {
...
+        name = this.translate_.getString('Point');
+      }
+      break;
+    case 'LineString':
+      name = this.translate_.getString('LineString');
+      break;
+    case 'Polygon':
+      name = this.translate_.getString('Polygon');
+      break;
+  }
```
pgiraud: Couldn't this be simplified to the following?
```
var name;
var type = feature.getGeometry().getType();
if (type == 'Point' && this.drawLabel.getActive()) {
  name = this.translate_.getString('Label');
} else {
  name = this.translate_.getString(type);
}
```

petzlux: The explicit listing of the feature types has more to do with allowing transifex to collect the strings for the different geom types .
with `this.translate_.getString(type)` gettext wont be able to collect all the possible geom type strings.

pgiraud: Good point!
Thanks for the explanation.

petzlux:
Okay addressed code review , will merge once Travis is happy

PR #959, opened 2015-12-02
Fix - Choose Map when no Map is owned by User

### LINKED ISSUE #919: Open a map when no map has been created yet
If the user never created any mymaps map and clicks on the "Open a map" button, nothing happens. The modal window doesn't show up.

Expected: the window should be open telling that there are no map available.

petzlux:
fix #919

### DIFF
--- a/geoportailv3/static/js/mymaps/mymapsdirective.js
+++ b/geoportailv3/static/js/mymaps/mymapsdirective.js
@@ -450,6 +450,10 @@ app.MymapsDirectiveController.prototype.openChooseMapModal = function() {
       } else if (!goog.array.isEmpty(mymaps)) {
         this.choosing = true;
         this.maps = mymaps;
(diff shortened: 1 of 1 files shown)


geoportailv3/static/js/mymaps/mymapsdirective.js
```diff
@@ -450,6 +450,10 @@ app.MymapsDirectiveController.prototype.openChooseMapModal = function() {
       } else if (!goog.array.isEmpty(mymaps)) {
         this.choosing = true;
         this.maps = mymaps;
+      } else if (goog.array.isEmpty(mymaps)) {
```
pgiraud: Is there an other case you are thinking of?
If not, I would use `else` with no condition.

petzlux:
Code review complete will merge once travis is happy

PR #960, opened 2015-12-02
upload symbol & thumbnail for feature

rmichaelis:
@jaykayone could you please have a look here : http://devv3.geoportail.lu/renaud/


petzlux:
will review

jaykayone:
functionallity IS good.
styling needs work but in other PR

PR #966, opened 2015-12-08
Mymaps layers

rmichaelis:
Could you please review and at least check 

1) The interraction between the url parameters and mymaps.
2) Already existing layers when a new map is opened.
3) Could someone propose a better icon when layers change.

Testing env here http://devv3.geoportail.lu/renaud

### DIFF
--- a/geoportailv3/static/js/catalog/catalogdirective.js
+++ b/geoportailv3/static/js/catalog/catalogdirective.js
@@ -6,8 +6,7 @@
  *
  * Example:
  *
- * <app-catalog app-catalog-map="::mainCtrl.map"
- *              app-catalog-currenttheme="mainCtrl.currentTheme">
+ * <app-catalog app-catalog-map="::mainCtrl.map">
  * </app-catalog>
  *
  * Note the use of the one-time binding operator (::) in the map expression.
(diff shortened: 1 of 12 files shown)


rmichaelis:
Theme management is missing

petzlux:
Okay will review

petzlux:
Regarding url parameters, who wins? Seems Url parameters take precedence over MyMaps settings.

petzlux:
Right now, the only option is to accept and save changes in the layers and backgroundlayers setting for a given MyMap. Can we add a No option, which would reset to the MyMap setting?

rmichaelis:
Some remaining questions :
- Should the map_id parameter be saved in the local storage?
- When should I save the starting view ?(x,y, zoom)
- Should it be saved together with the layers ?
- Same quesiton for the theme ?

jaykayone:
> Some remaining questions :

> Should the map_id parameter be saved in the local storage?
NO!

> When should I save the starting view ?(x,y, zoom)
We decided to use the extent of the features.
So we do not really need to save it.
> Should it be saved together with the layers ?
Yes we can do this. As soon as we change the extent, the popup comes and we save on "OK"
> Same quesiton for the theme ?
Yes

petzlux:
@jaykayone @rmichaelis  this issue is still there when for example loading a mymap with a different bgLayer to the existing one. I think this is an issue, as the user can-t reset the MyMap to the state it was last saved in, without reloading the whole page.

> Right now, the only option is to accept and save changes in the layers and backgroundlayers setting for a given MyMap. Can we add a No option, which would reset to the MyMap setting?

petzlux:
charging layers from the MyMap PilzKarte test `1e443e9c37534313a92144ae9af36d7b` doesnt seem to work.

PR #970, opened 2015-12-10
Share a mymaps

### LINKED ISSUE #787: Permalien d'accès direct à la carte
Link to #782

rmichaelis:
Fixes #787

### DIFF
--- a/geoportailv3/static/js/catalog/catalogdirective.js
+++ b/geoportailv3/static/js/catalog/catalogdirective.js
@@ -6,8 +6,7 @@
  *
  * Example:
  *
- * <app-catalog app-catalog-map="::mainCtrl.map"
- *              app-catalog-currenttheme="mainCtrl.currentTheme">
+ * <app-catalog app-catalog-map="::mainCtrl.map">
  * </app-catalog>
  *
  * Note the use of the one-time binding operator (::) in the map expression.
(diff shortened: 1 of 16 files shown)


rmichaelis:
@petzlux 
As disussed, sharing a mymaps is now in the sharing tools, and no more in mymaps. 
Could you please review ?

petzlux:
@rmichaelis Not sure if this is connected to this pull request, but if the mymap loaded does not have a bgLayer set, than no background map gets loaded

````
{layers_visibility: "true,true,true", rating: 4, create_date: "2013-03-19 09:05:35", bg_layer: "",…}
bg_layer: ""
bg_opacity: 34
category_id: 999
create_date: "2013-03-19 09:05:35"
description: "Wann se moies fësche ginn juppeidi juppeida"
label: ""
layers: "arrets_bus,,"
layers_indices: "5,8,9"
layers_opacity: "1,1,1"
layers_visibility: "true,true,true"
public: false
rating: 4
rating_count: 1
selected_node: "None"
theme: "main"
title: "Dem Fränz säin Treppeltour zu Useldéng"
update_date: "2015-10-22 16:08:53"
user_login: "fk"
uuid: "5bdb79809cba46a98ebab6f1b13c9116"
x: 92959.3
y: 66933.6
zoom: 9
````

rmichaelis:
For my point of view, if no bg_layer is set, then no background layer is displayed ?
What do you expect  as behaviour?

petzlux:
Even the  ````blank```` bgLayer is saved as ````bgLayer:blank````  , not an empty string ... So this seems to me like a case from old mymaps that we either need to convert on the server side, or take into account when loading a mymap

petzlux:
please also see #971

petzlux:
LGTM other issue will be addressed separately

Issue #975, opened 2015-12-17
Reset mymaps settings


rmichaelis:
Add a No option, which would reset to the MyMap setting

Issue #865, opened 2015-09-11
Generalize the use of `@export` for controller instance properties


elemoine:
Recent versions of Closure Compiler have the `--export_local_property_definitions` flag that can be used to "export" instance properties using the `@export` annotation. We [use](https://github.com/camptocamp/ngeo/pull/216) that annotation in ngeo, and we've started to use it in the project as well, for example in the [app-mymaps directive](https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportailv3/static/js/mymaps/mymapsdirective.js). Using this annotation instead of the square bracket notation makes the code easier to read, so we should generalize it.

jaykayone:
@elemoine is this something you wanna do retroactively?

elemoine:
> @elemoine is this something you wanna do retroactively?

We could. It would be good to make the code consistent. But this is not mandatory.

petzlux:
:+1:

PR #977, opened 2015-12-17
reset mymaps

rmichaelis:
Please review

### DIFF
--- a/geoportailv3/static/js/mymapsservice.js
+++ b/geoportailv3/static/js/mymapsservice.js
@@ -470,9 +470,8 @@ app.Mymaps.prototype.loadFeatures_ = function() {
 
 /**
  * update the map with layers
- * @private
  */
-app.Mymaps.prototype.updateLayers_ = function() {
+app.Mymaps.prototype.updateLayers = function() {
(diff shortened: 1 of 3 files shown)


rmichaelis:
fixes #975

geoportailv3/static/js/mymapsservice.js
```diff
@@ -470,9 +470,8 @@ app.Mymaps.prototype.loadFeatures_ = function() {
 
 /**
  * update the map with layers
- * @private
  */
```
petzlux: is their not an `@export` statement missing?

rmichaelis: I don't think so. 
I don't want to prevent renaming the function. I just want having the function available outside the class. (not private)

petzlux:
LGTM!

geoportailv3/static/js/mymapsservice.js
petzlux: okay then, I thought in the past that we would standardise on that behaviour, but guess didnt understand it properly. https://github.com/Geoportail-Luxembourg/geoportailv3/issues/865

PR #984, opened 2015-12-21
Popup is now draggable

### LINKED ISSUE #886: Prettify Info Popup
it should be in the top right corner of the window
Info window: "x" icon too low
move popup for editing vertices

rmichaelis:
fixes #886

### DIFF
--- a/geoportailv3/static/js/draw/featurepopup.html
+++ b/geoportailv3/static/js/draw/featurepopup.html
@@ -1,7 +1,7 @@
 <div ng-show="!ctrl.editingAttributes && !ctrl.editingStyle && !ctrl.deletingFeature">
   <div class="feature-popup-heading">
-    <button href class="close" ng-click="ctrl.close()">&times;</button>
-    {{ ctrl.feature.get('name') }}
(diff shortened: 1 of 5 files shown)


rmichaelis:
Could you please review

petzlux:
After dragging a feature popup, it becomes very hard to close it, have to click multiple times close icon .

jaykayone:
can you put it on /renaud so that I can test too please?

petzlux:
Question for @pgiraud 
This is probably out of scope of this pull request, but what is the logic for having a Popup Directive in geoportailv3 ? I ask, because I can see that ngeo has a Popup Directive https://github.com/camptocamp/ngeo/blob/master/src/services/popup.js

Are there features missing in the ngeo directive that make it necessary to recreate a Popup here?

rmichaelis:
@jaykayone done
@petzlux you are right the closing button should not be used to drag the popup.

jaykayone:
same remark for close button but for the rest it is OK

rmichaelis:
Is it better now ?

PR #995, opened 2016-01-05
Delete vertice using SHIFT key

### LINKED ISSUE #989: delete vertex
no possibility yet to delete vertex?

rmichaelis:
fixes #989

### DIFF
--- a/geoportailv3/static/js/draw/drawdirective.js
+++ b/geoportailv3/static/js/draw/drawdirective.js
@@ -324,11 +324,7 @@ app.DrawController = function($scope, ngeoDecorateInteraction,
       }, true, this);
 
   this.drawnFeatures_.modifyInteraction = new ol.interaction.Modify({
-    features: appSelectedFeatures,
-    deleteCondition: function(event) {
-      return ol.events.condition.shiftKeyOnly(event) &&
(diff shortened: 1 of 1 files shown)


rmichaelis:
@jaykayone is SHIFT key okay for you  ?

jaykayone:
after doing some user tests, the conclusion is:

there should be two possibilities:
- "Click on point and hit DELETE"
- Right click on point and click "DELETE" in menu (cf google mymaps)

pgiraud:
By default in OL3, if you single click on the vertex it is removed. This option has been deactivated in geoportailV3.
Isn't it a good solution before we get a contextmenu?

jaykayone:
yes this can be a good compromise for not to be blocked

pgiraud:
In this case, @rmichaelis you should completely remove the `deleteCondition`.

pgiraud:
FYI, I started to give the contextmenu a try in an isolated OL3 example.

rmichaelis:
Unfortunately, It's not so easy :-)
Removing the ```deletecondition``` throws unwanted event. 
The modify event is directly thraws after the create event.
That's the reason why we explicitly deactivated the default behaviour.

jaykayone:
+1 for the example

PR #999, opened 2016-01-07
export the feature as GPX and KML

rmichaelis:
@petzlux could you please review ?

### DIFF
--- a/geoportailv3/static/js/draw/featurepopupservice.js
+++ b/geoportailv3/static/js/draw/featurepopupservice.js
@@ -164,14 +164,14 @@ app.FeaturePopup.prototype.show = function(feature, map, opt_anchor) {
     this.setDraggable(angular.element(element));
   }, this);
   // Enable the dropdown menu
-  this.element_.find('[data-toggle=dropdown]')['dropdown']();
+  this.element_.find('[data-toggle=dropdown]').dropdown();
 };
 
 
 /**
  */
 app.FeaturePopup.prototype.toggleDropdown = function() {
-  this.element_.find('[data-toggle=dropdown]')['dropdown']('toggle');
+  this.element_.find('[data-toggle=dropdown]').dropdown('toggle');
 };
 
 
--- a/geoportailv3/static/js/draw/featurepopupservice.js
+++ b/geoportailv3/static/js/draw/featurepopupservice.js
@@ -79,6 +79,8 @@ app.FeaturePopup = function($compile, $rootScope, $document) {
   this.element_ = angular.element('<div app-feature-popup></div>');
   this.element_.addClass('feature-popup');
   this.element_.attr('app-feature-popup-feature', 'feature');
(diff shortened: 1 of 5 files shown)


geoportailv3/static/js/draw/featurepopupservice.js
```diff
@@ -159,6 +163,8 @@ app.FeaturePopup.prototype.show = function(feature, opt_anchor) {
   goog.array.forEach(headers, function(element) {
     this.setDraggable(angular.element(element));
   }, this);
+  // Enable the dropdown menu
+  $('.dropdown-toggle')['dropdown']();
```
rmichaelis: @pgiraud I would like to have your advice about that point.
As the popup element is dynamically inserted, the dropdown menu inside the popup is not automatically initialized.
That means the dropdown menu doesn't open.
I found this way to bypass the issue.

Is there a better alternative ?

pgiraud: In the following example, I demonstrate that there's no need to call 'dropdown' on elements added via jQuery and an ol.Overlay.
http://jsfiddle.net/pgiraud/36xvpng4/1/

I need to test what happens when the element is created with angular though.

rmichaelis: It probably depends on if bootsrap is already initialized when the element is created or not.

pgiraud: Actually, the reason why it doesn't work is because we use `stopEvent` for the overlay.
Calling `dropdown` on the element manually after the overlay is created seem to let new events to propagate through the map.

However, I would restrict this to the dropdowns in the overlay:
```javascript
this.element_.find('[data-toggle=dropdown]')['dropdown']();
```

Finally, it would be much pretty if `dropdown` was exposed as an extern in ngeo. Do you want me to create a pull request?

rmichaelis:
Yes please, create a pull request.

geoportailv3/static/js/draw/featurepopupservice.js
rmichaelis: Yes please, create a pull request.

petzlux:
@rmichaelis  shall i wait with review until issue is resolved ?

rmichaelis:
@petzlux You can already test the exported files.

petzlux:
the dropdown doesnt close afer clicking a menu item.

otherwise both exports work!

geoportailv3/static/js/draw/featurepopupservice.js
rmichaelis: @pgiraud don't spend too many time with this one. I'll do the job.

pgiraud:
https://github.com/camptocamp/ngeo/pull/503 has been merged.
You should now be able to use `.dropdown`.

rmichaelis:
It works. Thanks

PR #1021, opened 2016-01-19
Fix continue line

### LINKED ISSUE #958: Continue to draw a line is not available
After the line creation, it's no more possible to continue to draw the line

rmichaelis:
It fixes #958


rmichaelis:
@jaykayone Could you please test on /renaud and tell me if it's working as you are expecting

jaykayone:
works great even on tablet

but there seems to be one bug, at least on tablet.
After having continued a line and finished the drawing, I cannot select the line anymore in the map. I have to go through the list on the left to get into edit modus again

(tested with anonymous drawing)

rmichaelis:
@jaykayone It seems to be related with this : https://github.com/openlayers/ol3/issues/2105

PR #1046, opened 2016-01-28
Profile in mymaps public layers

### LINKED ISSUE #1039: MyMaps public layers , include profile in info tab

petzlux:
fix #1039


petzlux:
Ready for review @rmichaelis

petzlux:
@rmichaelis The ordering algorithm for multilinestrings in the python controller , as copied from geoportailv2 , doesnt return same results ... could you check to see why the resulting profiles are not consistent ?

rmichaelis:
@petzlux Could you please provide an example ?

PR #1064, opened 2016-02-19
Find and display the nearest address on right mouse click

### LINKED ISSUE #685: Reverse geocoding for location info

rmichaelis:
Fixes #685


geoportailv3/static/js/geocodingservice.js
```diff
@@ -0,0 +1,58 @@
+/**
+ * @fileoverview This file provides an Angular service for interacting
+ * with the "elevation" web service.
```
petzlux: tiny comment but we should probably correct this to read geocoding ... :-)

PR #1180, opened 2016-03-24
Set Piwik Site ID to theme

petzlux:
Record a visit using the right Piwik Site ID. This will also record a separate visit if user changes the Theme.

### DIFF
--- a/geoportailv3/static/js/themeservice.js
+++ b/geoportailv3/static/js/themeservice.js
@@ -7,16 +7,42 @@ goog.require('app.Themes');
 goog.require('ngeo.Location');
 
 
+/**
+ * @typedef {{push: function(Array<string>)}}
+ */
+app.Piwik;
+
+
 
 /**
  * @constructor
+ * @param {angular.$window} $window Global Scope.
  * @param {ngeo.Location} ngeoLocation ngeo Location service.
  * @param {app.Themes} appThemes
  * @param {app.ScalesService} appScalesService Service returning scales.
  * @param {Array.<number>} maxExtent Constraining extent.
  * @ngInject
  */
-app.Theme = function(ngeoLocation, appThemes, appScalesService, maxExtent) {
+app.Theme = function($window, ngeoLocation, appThemes,
+    appScalesService, maxExtent) {
+
+  /**
+   * @const
+   * @private
+   * @type {Object}
+   */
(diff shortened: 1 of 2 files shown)


petzlux:
@rmichaelis review and merge please!

geoportailv3/static/js/themeservice.js
```diff
@@ -7,16 +7,42 @@ goog.require('app.Themes');
...
+  /**
+   * @const
+   * @private
+   * @type {Object}
+   */
+  this.piwikSiteIdLookup_ = {
+    'main': 18,
+    'tourisme': 7,
+    'emwelt': 8,
+    'eau': 6
```
rmichaelis: Don't we need a different id for tourism V2 and V3 and for the other sites ? (As we do for the main site)

geoportailv3/static/js/themeservice.js
```diff
@@ -57,6 +83,15 @@ app.Theme = function(ngeoLocation, appThemes, appScalesService, maxExtent) {
  */
 app.Theme.prototype.setCurrentTheme = function(themeId, map) {
   this.currentTheme_ = themeId;
+
+  var piwikSiteId = this.piwikSiteIdLookup_[this.currentTheme_];
+  if (!goog.isDefAndNotNull(piwikSiteId)) {
+    piwikSiteId = this.piwikSiteIdLookup_[app.Theme.DEFAULT_THEME_];
```
rmichaelis: If the theme id is not found, we use the main theme.
Maybe would it be better to have a different site id ?

geoportailv3/static/js/themeservice.js
rmichaelis: Could it be useful to already have all the existing themes defined here ?

rmichaelis:
It works as expected.
Only a few non blocking comments.
You can merge if they are not relevant.

geoportailv3/static/js/themeservice.js
jaykayone: for the different ids: we wont keep tourismev2 so no need

for the complete list i agree but we can also add it later

geoportailv3/static/js/themeservice.js
jaykayone: has been like that for v2 so ok for me for now

PR #1230, opened 2016-04-27
PAG report generation

jaykayone:
Please Review @rmichaelis


rmichaelis:
The messages are now translated.
PR looks good to me.
Be careful : 
- Due to the new python libraries, the deployment needs a ```make clean all``` 
- The new messages need to be translated.

jaykayone:
OK for me when travis passes

PR #1290, opened 2016-05-24
Symbol of type point

rmichaelis:


rmichaelis:
@jaykayone Could you please check on http://devv3.geoportail.lu/renaud
Mymaps db should be updated

PR #1327, opened 2016-06-14
Show marker

pgiraud:
Adds a `showMarker`.
There may still need work on the `position` part about coordinate reference system.
@tonio Can you please review?

### DIFF
--- a/geoportailv3/static/less/geoportailv3.api.less
+++ b/geoportailv3/static/less/geoportailv3.api.less
@@ -1,1 +1,69 @@
 @import (inline) '../../../node_modules/openlayers/css/ol.css';
+@import "../../../node_modules/bootstrap/less/mixins/vendor-prefixes.less";
+
+
+@popup-bg: #fff;
+@popup-border-color: rgba(0,0,0,0.2);
+@popup-arrow-outer-color: @popup-border-color;
+@popup-arrow-color: @popup-bg;
+@popup-arrow-width: 11px;
+.lux-popup {
+  background-color: @popup-bg;
+  background-clip: padding-box;
+  border: 1px solid @popup-border-color;
+  border-radius: 5px;
+  .box-shadow(0 5px 10px rgba(0,0,0,.2));
+  z-index: 2;
+}
+
+.lux-popup-arrow {
+  border-width: @popup-arrow-width;
+  left: 50%;
+  margin-left: -@popup-arrow-width;
+  border-bottom-width: 0;
+  border-top-color: @popup-arrow-color;
+  bottom: -@popup-arrow-width;
+
+  &,
+  &:after {
+    position: absolute;
+    display: block;
+    width: 0;
+    height: 0;
+    border-color: transparent;
+    border-style: solid;
+  }
+
+  &:after {
+    border-width: @popup-arrow-width;
+    content: "";
+    bottom: 1px;
+    margin-left: -@popup-arrow-width;
+    border-bottom-width: 0;
+    border-top-color: @popup-arrow-color;
+  }
+}
+
+.lux-popup-content {
+  padding: 9px 14px;
+}
+
+.lux-popup-close {
+  -webkit-appearance: none;
+  padding: 0;
+  cursor: pointer;
+  background: 0 0;
+  border: 0;
+
+  float: right;
+  font-size: 21px;
+  font-weight: 700;
+  line-height: 1;
+  text-shadow: 0 1px 0 #fff;
+  opacity: .2;
+
+  &:hover {
+    text-decoration: none;
+    opacity: 0.5;
+  }
+}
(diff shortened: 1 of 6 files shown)


jsapi/src/main.js
```diff
@@ -26,3 +29,95 @@ lux.Map = function(options) {
...
+          position: position,
+          positioning: 'bottom-center',
+          offset: [0, -20],
+          insertFirst: false
+        });
+        this.addOverlay(popup);
+        this.renderSync();
+
+        if (!options.hover) {
+          var closeBtn = goog.dom.getElementsByClass('close', element)[0];
```
tonio: OL now uses `querySelectorAll`

jsapi/src/main.js
```diff
@@ -26,3 +29,95 @@ lux.Map = function(options) {
...
+  if (options.autoCenter) {
+    this.getView().setCenter(position);
+  }
+
+  if (options.html) {
+    var popup;
+    var showPopupEvent = options.hover ?
+        ol.events.EventType.MOUSEMOVE : ol.events.EventType.CLICK;
+    ol.events.listen(element, showPopupEvent, (function() {
+      if (!popup) {
```
tonio: Missing else to re-add existing popup?

jsapi/src/main.js
```diff
@@ -26,3 +29,95 @@ lux.Map = function(options) {
...
+        var element = buildPopupLayout_(options.html, !options.hover);
+        document.body.appendChild(element);
```
tonio: Is this necessary?

geoportailv3/static/less/geoportailv3.api.less
```diff
@@ -1 +1,77 @@
 @import (inline) '../../../node_modules/openlayers/css/ol.css';
+@import "../../../node_modules/bootstrap/less/mixins/vendor-prefixes.less";
+
+
+@popup-bg: #fff;
+@popup-border-color: rgba(0,0,0,0.2);
+@popup-arrow-outer-color: @popup-border-color;
+@popup-arrow-color: @popup-bg;
+@popup-arrow-outer-width: 11px;
+@popup-arrow-width: 11px;
```
tonio: Do we need 2 variables?

geoportailv3/static/less/geoportailv3.api.less
```diff
@@ -1 +1,77 @@
...
+      border-bottom-width: 0;
+      border-top-color: @popup-arrow-color;
+    }
+  }
+
+  .lux-popup-content {
+    padding: 9px 14px;
+  }
+
+  .close {
```
tonio: I’ll add prefix to avoid name collisions.

geoportailv3/static/less/geoportailv3.api.less
```diff
@@ -1 +1,77 @@
...
+@popup-arrow-width: 11px;
+.lux-popup {
+  background-color: @popup-bg;
+  background-clip: padding-box;
+  border: 1px solid @popup-border-color;
+  border-radius: 5px;
+  .box-shadow(0 5px 10px rgba(0,0,0,.2));
+  z-index: 2;
+
+  & > .lux-popup-arrow {
```
tonio: With prefix I don’t think nesting is necessary?

jsapi/src/main.js
pgiraud: Actually, the popup is (supposed to be) destroyed each time it's closed. Now I see that it would be better not to recreate i but to re-add it instead.

pgiraud: Modified.

jsapi/src/main.js
pgiraud: It's not. Good catch.

geoportailv3/static/less/geoportailv3.api.less
pgiraud: I reused the same rules/variables as bootstrap. We indeed don't need so much complexity.

geoportailv3/static/less/geoportailv3.api.less
pgiraud: Nesting simplified.
I also removed the 'lux-popup-top' rule since there's only one possible option for positioning.

geoportailv3/static/less/geoportailv3.api.less
```diff
@@ -1 +1,71 @@
...
+  background: 0 0;
+  border: 0;
+
+  float: right;
+  font-size: 21px;
+  font-weight: 700;
+  line-height: 1;
+  text-shadow: 0 1px 0 #fff;
+  filter: alpha(opacity=20);
+  opacity: .2;
```
tonio: why having `filter` + `opacity` ?

tonio:
One minor comment, otherwise LGTM.

pgiraud:
Travis build fails. But this is not related to the proposed changes.

PR #1375, opened 2016-07-12
Use remote webfont now that CORS are fixed

tonio:


pgiraud:
Looks good to me. Please merge.

PR #1397, opened 2016-08-02
fix sentry_dsn setting per instance

petzlux:


petzlux:
@jaykayone you can get the dsn from the commit

just add to vars_main the following var : `sentry_dsn`

PR #1404, opened 2016-08-04
Add support for external vector layers

pgiraud:
This pull request adds support for GPX and KML vector files.
The user can call `addGPX` and `addKML` on the `lux.Map` object. The map view fits automatically (center and zoom) to the features added.
Clicking on the vector features, a popup is displayed with the attributes for the given features.

Please review.


jsapi/src/main.js
```diff
@@ -437,6 +450,138 @@ lux.Map.prototype.addSearch = function(target) {
...
+  if (opt_styleFunction) {
+    options.style = opt_styleFunction;
+  }
+  this.promise.then(function() {
+    var vector = new ol.layer.Vector(options);
+    this.addLayer(vector);
+
+    ol.events.listen(vector.getSource(), ol.source.VectorEventType.ADDFEATURE,
+        function() {
+          var size = /** {ol.Size} */ (this.getSize());
```
tonio: Is the typecast mandatory ? (looks suspicious w/o the `@type`)

jsapi/src/main.js
```diff
@@ -437,6 +450,138 @@ lux.Map.prototype.addSearch = function(target) {
...
+          goog.asserts.assert(size !== undefined, 'size should be defined');
```
tonio: why an assert and not just a if?

jsapi/src/main.js
```diff
@@ -437,6 +450,138 @@ lux.Map.prototype.addSearch = function(target) {
...
+      var properties = feature.getProperties();
+
+      var html = '<table>';
+      var key;
+      for (key in properties) {
+        if (key != feature.getGeometryName()) {
+          html += '<tr><th>';
+          html += key;
+          html += '</th><td>';
+          html += properties[key] + '</tr>';
```
tonio: minor: missing closing `td`

jsapi/src/main.js
```diff
@@ -437,6 +450,138 @@ lux.Map.prototype.addSearch = function(target) {
...
+      if (!features.getLength()) {
+        return;
+      }
+      var feature = features.getArray()[0];
+      var properties = feature.getProperties();
+
+      var html = '<table>';
+      var key;
+      for (key in properties) {
+        if (key != feature.getGeometryName()) {
```
tonio: Maybe we should test if `properties[key]` is not null or empty string?

tonio:
screenshot from 2016-08-24 09-34-19

I guess we should:
 - use the same font as other controls,
 - deal with text overflows (set value as title in case of hidden overflow?)

jsapi/src/main.js
pgiraud: It can help the integrator to better understand the reason why the extent is not changed.

pgiraud:
Thanks for the reviews.

PR #1432, opened 2016-09-15
Less intrusive CSS for controls

pgiraud:
Make it so styling for mouse position or background selector controls is applied only when added to the map.


jaykayone:
@pgiraud works OK but there is one detail: the search box in a div shows the "x" component outside of the searchfield. --> http://apps.geoportail.lu/exemple_api/apiv3_demo.html#example7

PR #1444, opened 2016-09-19
Resize the map when the side panel is opened

rmichaelis:


pgiraud:
@rmichaelis Can you please explain the changes? It would be nice to get this back to ngeo.

rmichaelis:
We finally found a computer where the map is distorted at the opening, and we can easily reproduce the issue.

As first try, Eric tried to solve this issue using a kind of timeout, before forcing a call to the updatesize function.

It appeared, that the timeout was not big enough, and I increased it up to 2 seconds.
Most of the problems seemed to be solved, but some client still complained.

Thus, Instead of increasing it once again, I choose to listen on the "transitionend" event. That means, when the transition has been done, and the side panel is completely opened(or closed), the map is refreshed.

I think that the code with animationDelay is no more necessary. 
But I fear to cause side effects by removing this part of code...

geoportailv3/static/js/resizemap.js
```diff
@@ -47,14 +46,19 @@ app.resizemapDirective = function($window, $animate) {
...
+                } else {
+                  map.updateSize();
+                  map.renderSync();
                 }
               }, $window);
 
+          element.bind('transitionend',function() {
+            map.updateSize();
+            map.renderSync();
+          }.bind(this));
```
pgiraud: No need for bind here.

pgiraud:
@rmichaelis I applied your changes to the `ngeo` `resizemap` directive. I'm afraid the changes in the `animationDelay` callback break something. The map doesn't resize any more during the animation.
See https://pgiraud.github.io/ngeo/resize/examples/animation.html

In this second example, I kept the code as is in `animationDelay`.
https://pgiraud.github.io/ngeo/resize2/examples/animation.html

@rmichaelis I'd be happy if you could confirm that this second example works in the computer on which you had problems. Please also confirm that the problem appears on the master branch on this computer.
https://camptocamp.github.io/ngeo/master/examples/animation.html

rmichaelis:
@pgiraud This url https://pgiraud.github.io/ngeo/resize/examples/animation.html returns a 404 not found.

rmichaelis:
@pgiraud Yes you are right, I broke something. In fact, I pushed the wrong commit. It's now OK.

The second example as well as the third work on the computer.

I've the feeling that the problem is linked with the initialization time of the Geoportal. 
As your examples are "light", then the problem does not appear.

geoportallux:
peux-tu d?ployer?

Jeff Konnen

Charg? d'?tudes - informaticien principal/Chef de projet

LE GOUVERNEMENT DU GRAND-DUCH? DE LUXEMBOURG
Minist?re des Finances
Administration du cadastre et de la topographie
Service informatique & Service g?oportail

54, avenue Gaston Diderich<x-apple-data-detectors://1/0> . L-1420 Luxembourg
T?l. (+352) 44901-261<tel:(+352)%2044901-261> . Fax (+352) 44901-333<tel:(+352)%2044901-333>
E-mail: jeff.konnen@act.etat.lu<mailto:jeff.konnen@act.etat.lu>

www.act.public.lu<http://www.act.public.lu/> www.geoportail.lu<http://www.geoportail.lu/>

On 21.09.2016, at 09:21, Renaud Micha?lis <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #1444<https://github.com/Geoportail-Luxembourg/geoportailv3/pull/1444>.

-

Issue #398, opened 2015-04-09
ngeoCreatePopup avec différents templates


rmichaelis:
I would like to use ngeoCreatePopup to display information about the layers. 
In the other hand, a value for ngeoPopupTemplateUrl is already defined to display metadata information. 

Is it possible to use ngeoCreatePopup  twice with a different template ? 
Could you please explain me how I could achieve it ?

Note that it's only for my understanding and for test purpose. Jeff  just remembered me that the information has to be display in the info tab panel.

Thanks,

pgiraud:
This is indeed impossible for now.
Please don't hesitate to tell us if this is eventually required to handle different templates.

elemoine:
Is this still relevant?

rmichaelis:
No more relevant. That was just for testing the popup directive. Thanks.

PR #1454, opened 2016-09-26
Feedback tool

### LINKED ISSUE #1106: Outil de reporting
créer un outil de feedback des utilisateurs cf outil suisse

petzlux:
fix #1106
@rmichaelis 
Can you help me? I dont understand how to change the template for the popup from the one set for the [layerinfo template url](https://github.com/Geoportail-Luxembourg/geoportailv3/blob/3fd6f5c300914889cfb0210bc8388ab132c2b11e/geoportailv3/templates/index.html#L287)

According to [ngeo source code](https://github.com/camptocamp/ngeo/blob/87cdea826e091a7447e41f36fd2ce3fe0e54701f/src/directives/popup.js#L28) need to override the `ngeoPopupTemplateUrl` but didnt manage to do this successfully.

### DIFF
--- a/geoportailv3/static/js/feedback/feedback.html
+++ b/geoportailv3/static/js/feedback/feedback.html
@@ -1,0 +1,38 @@
+<div class="feedback-content">
+  <h3 translate>Report a map content problem:</h3>
+  <form ng-submit="ctrl.sendReport()">
+    <div class="form-group">
+      <label for="inputEmail" translate>Your email</label>
+      <input type="email" required ng-model="ctrl.email" class="form-control" id="inputEmail">
+    </div>
+    <div class="form-group">
+      <label for="dropdownLayerPicker" translate>Pick layer where problem occurs</label>
+      <div class="dropdown" id="dropdownLayerPicker">
+        <button type="button" class="btn btn-default" data-toggle="dropdown">
+          {{ctrl.concernedLayer | translate}}<span class="caret"></span>
+        </button>
+        <ul class="dropdown-menu" role="menu">
+         <li>
+            <a href ng-click="ctrl.setConcernedLayer(ctrl.bgLayer)">{{ctrl.bgLayer.get('label')|translate}}</a>
+          </li>
+          <li ng-repeat="layer in ctrl.layers">
+            <a href ng-click="ctrl.setConcernedLayer(layer)">{{layer.get('label')|translate}}</a>
+          </li>
+        </ul>
+      </div>
+    </div>
+    <div class="form-group">
+      <label for="inputDescription" translate>Describe the problem within the current map extent:</label>
+      <textarea rows="5" required  maxlength="500" ng-model="ctrl.description" class="form-control" id="inputDescription"></textarea>
+    </div>
+    <!-- <div class="form-group"> -->
+    <!--   <label for="exampleInputFile">Attach a file (optional)</label> -->
+    <!--   <input type="file" id="exampleInputFile"> -->
+    <!-- </div> -->
+    <div class="form-group">
+      <p translate>The following <a ng-href="{{ctrl.url}}" target="_blank">map link</a> will be sent to our support team.</p>
+      <p translate>You can also use the <a href ng-click="ctrl.activateDrawingTools()">Drawing Tools</a> to annotate the map send to the support team.<p></p>
+    </div>
+    <button type="submit" class="pull-right btn btn-default" translate>Send feedback to support team</button>
+  </form>
+</div>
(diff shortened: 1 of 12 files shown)


rmichaelis:
@petzlux Unforunately I don't think this is possible.
I made some tests in the past and here is the conclusion : https://github.com/Geoportail-Luxembourg/geoportailv3/issues/398

petzlux:
@jaykayone two big issues here we need to discuss: 

1. At the moment, the ngeo popup component only allows for one global template for content per project. This means that we will have to rewrite Layer Metadata popup template to be generic so to be reusable for the feedback tool popup.
2. The ngeo popup component is not draggable. This will be a big UX problem when users want to report an issue that will be hidden by the popup.

Can we ask for both problems CamptoCamp to improve the ngeo popup ?

rmichaelis:
Could we use what we did for mymaps ?

rmichaelis:
image

jaykayone:
good point, or directly use the popup used in geoadmin?

petzlux:
@jaykayone Can we decide to do the feedback tool in the sidebar?

jaykayone:
yes

geoportailv3/views/feedback.py
```diff
@@ -0,0 +1,47 @@
...
+                "<h3>La couche suivante est concernee:</h3>" \
+                "<p>{2}</p>" \
+                .format(vars['email'],
+                        sanitized_description,
+                        vars['layer'],
+                        vars['url']
+                        )
+            message = Message(
+                author=vars['email'],
+                to='weber.patrick@act.etat.lu',
```
rmichaelis: It could be useful to get this from the config file.

petzlux: fixed

rmichaelis:
Clicking the "draw" button does not open the mymaps tab anymore.

petzlux:
That is by design, otherwise the feedback tab would be hidden.

rmichaelis:
You should add the following key support_email_address: support@geoportail.lu
in file vars_geoportailv3.yaml to avoid problems

rmichaelis:
I would say that clicking the "draw" button should not open the mymaps tab only when you are in the feedback tool. Otherwise it should open it.

petzlux:
That should be the behaviour! have you tried!?

rmichaelis:
Yes I tried and this is not working as expected.

rmichaelis:
Steps to reproduce : 
1. Open the feedback tool
2. Open the mymaps tab
3. Draw an object
4. Resize your browser
Then the feedback tab is displayed below the map

image

petzlux:
Yes I struggled to get this working, as the sidebar is managed by the button group component (layers,mymaps,info,legends) but the feedback is not part of that... . will see if i can fix that

petzlux:
@rmichaelis I fixed the problem with the sidebar panel ,and the listener should also get removed once the feedback tool is not visible anymore. 

can you please check.

petzlux:
also now available for review under http://beta.geoportail.lu

geoportailv3/static/js/feedback/feedback.html
```diff
@@ -0,0 +1,38 @@
...
+    <div class="form-group">
+      <label for="inputDescription" translate>Describe the problem within the current map extent:</label>
+      <textarea rows="5" required  maxlength="500" ng-model="ctrl.description" class="form-control" id="inputDescription"></textarea>
+    </div>
+    <!-- <div class="form-group"> -->
+    <!--   <label for="exampleInputFile">Attach a file (optional)</label> -->
+    <!--   <input type="file" id="exampleInputFile"> -->
+    <!-- </div> -->
+    <div class="form-group">
+      <p translate>The following <a ng-href="{{ctrl.url}}">map link</a> will be sent to our support team.</p>
```
rmichaelis: Maybe should you add a target="_blank" in this link, because if the user clicks on it , he could loose the information  that he already wrote.

rmichaelis:
Just one comment you may want to address. Otherwise it looks good to me.

petzlux:
LGTM ?!?

geoportailv3/static/js/feedback/feedback.html
petzlux: done

jaykayone:
Should picking a layer be mandatory?

geoportallux:
Hmm background layers are also layers. Enough to be used for a feedback ?

Francis Kaell
Ingénieur-géomètre officiel / responsable du géoportail
LE GOUVERNEMENT DU GRAND-DUCHÉ DE LUXEMBOURG
Ministère des Finances
Administration du cadastre et de la topographie
Service géoportail
54, avenue Gaston Diderich . L- 1420 Luxembourg
Tél. (+352) 44901 - 244 Fax (+352) 44901 - 333
E-Mail: francis.kaell@act.etat.lu<mailto:francis.kaell@act.etat.lu>
www.act.public.lu<http://www.act.public.lu/>    www.geoportail.lu<http://www.geoportail.lu/>

[siginspiring]

PR #1458, opened 2016-09-28
Search highlight

### LINKED ISSUE #1452: API: highlight geometries after searching
for addresses, we should show the point on the map when selecting a search result (just like we do in the app)

pgiraud:
This pull request makes sure that the found address is highlighted on the map.

Also, we now make sure that only addresses are searched for.

Fixes #1452.

Please review.

### DIFF
--- a/jsapi/src/main.js
+++ b/jsapi/src/main.js
@@ -708,7 +708,7 @@ lux.Map.prototype.addSearch = function(target) {
     clear.style.display = (input.value == '') ? '' : 'block';
   });
 
-    var fillStyle = new ol.style.Fill({
+  var fillStyle = new ol.style.Fill({
     color: [255, 255, 0, 0.6]
   });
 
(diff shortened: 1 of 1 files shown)


jsapi/src/main.js
```diff
@@ -708,7 +741,7 @@ lux.Map.prototype.addSearch = function(target) {
     'menuClass' : 'lux-search-suggestions',
     'source'    : function(term, suggest) {
       term = term.toLowerCase();
-      fetch(lux.searchUrl + 'limit=5&query=' + term).then(function(resp) {
+      fetch(lux.searchUrl + 'limit=5&layer=Adresse&query=' + term).then(function(resp) {
```
jaykayone: I think the integrator should have the possibility to override this setting.
As a default it is Ok, but the integrator should be able to define his own search target. @petzlux @rmichaelis what do you think?

rmichaelis: Yes at least the limit should be overridable. 
The layer I'm not sure ?

pgiraud: In the initial offer it was asked that the in the API the search should be limited to addresses.
Let's open a different issue if it is expected to support more types.

PR #1472, opened 2016-10-03
Add support for reloadInterval for GPX/KML layers

### LINKED ISSUE #1455: reload strategy when adding KML / GPX
When adding a GPX / KML file to a map, the user should be able to define the reload strategy in the options

http://openlayers.org/en/v3.7.0/apidoc/ol.source.TileWMS.html#updateParams

cf: integrators wanting to choose a kml layer to be reloaded once a minute like here: http://cita.lu/

pgiraud:
Fixes #1455

### DIFF
--- a/jsapi/src/main.js
+++ b/jsapi/src/main.js
@@ -1,6 +1,7 @@
 goog.provide('lux');
 goog.provide('lux.Map');
 
+goog.require('goog.Uri');
 goog.require('goog.asserts');
 goog.require('goog.dom');
 goog.require('lux.LayerManager');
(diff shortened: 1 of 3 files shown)


jsapi/src/main.js
```diff
@@ -778,39 +780,62 @@ lux.Map.prototype.addGPX = function(url) {
...
 /**
  * Adds a KML file on the map
  * @param {string} url Url to the GPX file
+ * @param {luxx.VectorOptions=} opt_options Options.
  * @export
  */
-lux.Map.prototype.addKML = function(url) {
-  this.addVector(url, new ol.format.KML());
+lux.Map.prototype.addKML = function(url, opt_options) {
+  this.addVector(url, new ol.format.KML(), undefined, opt_options);
```
tonio: It may have been easier to read with a single object for options, rather than 2 parameters…

PR #1481, opened 2016-10-04
Highlight the shown features if corresponding layer is visible

### LINKED ISSUE #1479: showFeatures should highlight
http://apps.geoportail.lu/exemple_api/apiv3_demo.html#example11

the feature is recentered on but not highlighted. 

I would propose the following: 
If the layer is present, it should be highlighted, otherwise not. (in this example parcels is present so it should be highlighted. if parcels not present, only zoom recenter)

pgiraud:
Fixes #1479

### DIFF
--- a/jsapi/src/main.js
+++ b/jsapi/src/main.js
@@ -210,6 +210,41 @@ lux.Map = function(options) {
    */
   this.searchLayer_ = null;
 
+  var fillStyle = new ol.style.Fill({
+    color: [255, 255, 0, 0.6]
+  });
+
+  var strokeStyle = new ol.style.Stroke({
+    color: [255, 155, 55, 1],
+    width: 3
+  });
+
+  /**
+   * @private
+   * @type {ol.style.Style} The style to be used for the show and search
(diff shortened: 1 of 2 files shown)


jsapi/src/main.js
```diff
@@ -721,6 +766,8 @@ lux.Map.prototype.addFeature = function(json) {
     return;
   }
   var size = this.getSize();
+  // remove any highlighted feature
+  this.showLayer_.getSource().clear();
```
tonio: `addFeature` may be called several times (cf [foreach](https://github.com/Geoportail-Luxembourg/geoportailv3/pull/1481/files#diff-80f6c764111465dfaa94038b75e98be0L695)),  `clear` should maybe better done in `showFeatures`…

pgiraud: Fixed.

tonio: `git push` ? (-:

pgiraud: Pushed.

PR #1494, opened 2016-10-07
Update ngeo

rmichaelis:


jaykayone:
did you really want to commit the PO files?

by the way, I cannot test, the print server does not seem to work on the dev server

PR #1499, opened 2016-10-10
Set the base url only once

### LINKED ISSUE #1496: initUrls may be called several times
When several maps are added in a page the `initUrls` function may be called several times which results in wrong urls.

pgiraud:
Fixes #1496.
@tonio please review.

### DIFF
--- a/jsapi/src/main.js
+++ b/jsapi/src/main.js
@@ -43,9 +43,28 @@ lux.searchUrl = 'fulltextsearch?';
 lux.baseUrl = null;
 
 /**
- * @param {string} url Base url to services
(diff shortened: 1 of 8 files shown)


jsapi/src/main.js
```diff
@@ -43,9 +43,25 @@ lux.searchUrl = 'fulltextsearch?';
 lux.baseUrl = null;
 
 /**
- * @param {string} url Base url to services
+ * @param {string|null} url Base url to services
```
tonio: missing `@export` ?

pgiraud: Good catch.

PR #1521, opened 2016-10-13
Add support for query on map

### LINKED ISSUE #1338: Interrogation
Interrogation sur la carte + affichage d'un popover avec les attributs du/des features(s) sélectionné.

pgiraud:
Fixes #1338.

This will need to be tested in real cases.

### DIFF
--- a/jsapi/src/main.js
+++ b/jsapi/src/main.js
@@ -50,9 +50,8 @@ lux.setBaseUrl = function(url) {
   if (!url) {
     lux.layersUrl = '../layers.json';
     lux.i18nUrl = '../lang_xx.json';
-    lux.queryUrl = 'http://localhost:5000/getfeatureinfo?';
 
-    url = 'http://map.geoportail.lu/main/wsgi/';
+    url = 'http://apiv3.geoportail.lu/api/wsgi/';
   } else {
     lux.layersUrl = url + lux.layersUrl;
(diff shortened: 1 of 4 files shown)


jsapi/src/main.js
```diff
@@ -1195,3 +1206,92 @@ lux.reverseGeocode = function(coordinate, cb) {
...
+      positioning: 'bottom-center',
+      offset: [0, -20],
+      insertFirst: false
+    });
+
+    this.addOverlay(this.queryPopup_);
+    this.renderSync();
+
+    var closeBtn = element.querySelectorAll('.lux-popup-close')[0];
+    ol.events.listen(closeBtn, ol.events.EventType.CLICK, function() {
```
tonio: Shouldn’t this be in `buildPopupLayout` (with a closeCallback if needed)?

jaykayone:
I have not tested this yet, but is there also a possibility to send the content to a div instead of a popup ?

We did this for all the other components, it would be good to have that possibility too

jsapi/src/main.js
```diff
@@ -1090,6 +1099,7 @@ lux.WMTSLayerFactory_ = function(config, opacity) {
     name  : config['name'],
     id: config['id'],
     metadata: config['metadata'],
+    queryable_id: config['id'],
```
pgiraud: I'm using the same names as the main app.

PR #1525, opened 2016-10-14
Print PNG

### LINKED ISSUE #902: Export an image
In addition to the print as PDF, the print should also be available as png

rmichaelis:
Fixes #902 
@jaykayone Could you please check layout here : devv3.geoportail.lu/renaud


jaykayone:
layout is OK for me.
I would only grey out PNG when "with legend" is chosen

rmichaelis:
Done

PR #1595, opened 2016-11-29
remove vars_main from sourcecode

petzlux:


jaykayone:
attention when this gets merged and deployes we will lose our config

PR #1596, opened 2016-11-29
update of translations

petzlux:
regular reminder that we should fix our translations in the source code.


petzlux:
@jaykayone Can you check if these translations correspond to the prod version in use ?

petzlux:
@rmichaelis can you rebase this pull request on devv3 so that we have the right catalog.

PR #1615, opened 2016-12-13
Fix geolocation

### LINKED ISSUE #1535: geolocalisation
while waiting for https-only:

reload the site in https when people click on the geoloc button

rmichaelis:
Fixes #1535 
@jaykayone Could you please test redirect if not https ?
http://devv3.geoportail.lu/renaud


jaykayone:
it works but ideally the golocation should then be activated after the reload

PR #1621, opened 2016-12-14
Do not allow to start with a blank map

### LINKED ISSUE #1266: load default background
when no layer is loaded + message

rmichaelis:
Fixes #1266 
@jaykayone Could you please check on dev/renaud if it works as expected ?


jaykayone:
yes ok

PR #1622, opened 2016-12-19
Allow to draw and share with the mobile view

### LINKED ISSUE #709: mobile - sharing
there's no sharing button on mobile... is that what we want @petzlux ?

rmichaelis:
Fixes #709 
@jaykayone Could you please check on dev/renaud
Could you please check the user experience with mobile as well as the desktop view.


jaykayone:
there are borders missing for the icons

jaykayone:
drawing interaction on mobile is a bit weird.
especially the interaction between drawing and panning.
but it is totally usable

jaykayone:
profile is not updatimg after editing on mobile

PR #1635, opened 2017-01-12
Add current print date

### LINKED ISSUE #1288: Date on PDF print
add the generation date of the PDF

rmichaelis:
Fixes #1288


rmichaelis:
Default values are available. Then no more print crash.

@jaykayone Could you please test date on the left side here : http://devv3.geoportail.lu/main/

PR #1736, opened 2017-03-22
Update dependencies

fredj:
Update project dependencies (angular, OpenLayers and ngeo).

todo:
 - [x] Profile does not work anymore.
 - [x] Point Drawing : Selection of the "width" using the range selector does not allow to select the entire range.
 - [x] Unable to login despite the fact that the login webservice returns true.
 - [x] Layer information does not work.
 - [ ] check if the image upload works
 - [x] Use `constraintResolution` option in `MouseWheelZoom` and `PinchZoom`
 - [x] Fix tooltip styling (measure) -> add missing `ngeo-` prefix to css class
 - [x] Remove step attribute from input number (style editing)
 - [x] Print symbol
 - [ ] input number when the step is not 1


rmichaelis:
Profile does not work anymore.

rmichaelis:
Point Drawing : Selection of the "width" using the range selector does not allow to select the entire range.

rmichaelis:
Unable to login despite the fact that the login webservice returns true.

rmichaelis:
Layer information does not work.

fredj:
@rmichaelis the callback name should be fixed now

rmichaelis:
It works now. Thanks.

fredj:
@rmichaelis the login and the vector style update should be fixed now, can you please test?

rmichaelis:
It looks fine.

rmichaelis:
@fredj The last commit seems to solve the remaining issues.  Are you ready with this PR ? 
@jaykayone could you have a try on devv3/renaud  ?

jaykayone:
on macintosh I saw that when i scroll with the mouse, it zooms seemlessly between the fixed tile scales, which has the effect that the scale bar shows an empty value:

screen shot 2017-04-04 at 15 20 44

rmichaelis:
Background color of the area was better in the old version:
New measurement tool
image
Old measurement tool
image

rmichaelis:
Unable to write the number using the keyboard. 
The granularity should not be the same in the text area as in the selector.
image

fredj:
I've updated this ticket description with your remarks, you are very good testers BTW :-)

rmichaelis:
Printting a symbol does not work anymore.
"graphicFormat" seems to be lost in the symbolizers attribute.
image
image

fredj:
@rmichaelis for the print issue:
It's not working when an image symbol is used, and it's working when a shape (circle, square, triangle, ...) is used, can you confirm?

Maybe the issue is that the print can't access devv3/renaud (mapfish print dnsMatch config ?)

rmichaelis:
Yes its working with shape and not with symbols. 
You can test here : http://g-o.lu/3/RLaJ

The problem comes frome here :
https://github.com/camptocamp/ngeo/commit/a57a08364674732b4ca763c367ae43dda8965d7f

If the graphicFormat is not provided, map fish print is not working.
map fish print works fine even if the provided type is not the real one.

fredj:
@rmichaelis I've updated this PR, all the mentioned issues should be fixed now. Can you please update devv3/renaud and test again?

rmichaelis:
devv3/renaud is updated. It seems to be ok. Except this: Arrows are not working.
image

@jaykayone : could you please try ?

fredj:
The arrows are not working if the step of the corresponding slider is not 1

fredj:
The problem is that when the arrow is clicked, the value is incremented by 1 and the setter function is correctly called (50 to 51).

Then, the value is set to the slider but since the step is 5 instead of 1, the value is set back to the original (51 to 50).

We will have to remove the step from the slider or add the same step to the number input

rmichaelis:
Try to select a drawing on the map, returns this error : 

`querydirective.js:466 Uncaught TypeError: Cannot read property 'getArray' of undefined
    at querydirective.js:466
    at forEachFeatureAtCoordinate (map.js:120)
    at vectorlayer.js:199
    at replaygroup.js:234
    at ol.render.canvas.PolygonReplay.ol.render.canvas.Replay.replay_ (replay.js:419)
    at ol.render.canvas.PolygonReplay.ol.render.canvas.Replay.replayHitDetection (replay.js:568)
    at ol.render.canvas.ReplayGroup.replayHitDetection_ (replaygroup.js:369)
    at ol.render.canvas.ReplayGroup.forEachFeatureAtCoordinate (replaygroup.js:222)
    at ol.renderer.canvas.VectorLayer.forEachFeatureAtCoordinate (vectorlayer.js:189)
    at ol.renderer.canvas.Map.ol.renderer.Map.forEachFeatureAtCoordinate (map.js:147)`

fredj:
> Try to select a drawing on the map, returns this error ...

fixed with c8b2f3d

rmichaelis:
It works fine. Thanks.

rmichaelis:
@fredj We would like to merge this pull request.
Could you please rebase ? If needed, I'll solve the step problems by myself.

fredj:
@rmichaelis code rebased.
Feel free to merge, one solution (untested) to fix the step issue is to revert this commit: cbb63c3

rmichaelis:
@fredj Thanks for the work.

PR #1740, opened 2017-03-23
Fix Piwik not counting visits to prof theme

petzlux:


petzlux:
@rmichaelis review please?

PR #1746, opened 2017-03-28
Minor css fixes to poi-feature form

fredj:
This is a minor css enhancement to the poi-feature form:
 - better padding above the footer buttons
 - only allow vertical resize to the textarea

before:
draw_form

after:
draw_form_after


fredj:
Thanks for the review, I don't have write access to this repository

PR #1757, opened 2017-04-03
Search Enter key autocompletes first result

### LINKED ISSUE #1742: Wrong action when hitting enter in search box
When typing eg Schengen and then hitting Enter, it opens the metadata of "Typologie des eaux de surfaces".

Expected behaviour: Zoom to the first feature in the list
Alternative: Do nothing

petzlux:
fix #1742 enter autocompletes first result

another press on enter then selects the result


petzlux:
@rmichaelis review please

PR #1764, opened 2017-04-06
Fix layers2es for blank metadata entries

petzlux:


petzlux:
@jaykayone here is the hotfix as promised

PR #1818, opened 2017-06-08
Remove install of node

rmichaelis:


fredj:
See https://github.com/camptocamp/ngeo/issues/2676

Issue #1849, opened 2017-07-11
print text not aligned left


jaykayone:
texts in drawings are aligned left, in print they seem to be aligned at centre

PR #1875, opened 2017-08-31
Same alignment both in the print and on the screen

rmichaelis:
In ngeo print service when a text rotation is provided, then text alignment is set to center.  Text aligment in Geoportail is left. https://github.com/camptocamp/ngeo/blob/master/src/services/print.js#L755


rmichaelis:
It Fixes #1849

PR #1910, opened 2017-11-06
get print servers info from db

jaykayone:
@rmichaelis please review

### DIFF
--- a/geoportailv3/views/luxprintproxy.py
+++ b/geoportailv3/views/luxprintproxy.py
@@ -81,6 +81,14 @@ class LuxPrintProxy(PrintProxy):
         print_servers = DBSession.query(LuxPrintServers).all()
         print_urls = [[print_server.url] for print_server in print_servers]
         if print_urls is not None and len(print_urls) > 0:
+            
+            for url in print_urls:
+                try:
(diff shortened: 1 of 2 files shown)


geoportailv3/views/luxprintproxy.py
```diff
@@ -78,9 +79,23 @@ class LuxPrintProxy(PrintProxy):
     @view_config(route_name="lux_printproxy_report_create")
     def lux_report_create(self):
         token = self.config["authtkt_secret"]
-        print_urls = self.config["print_urls"]
+        print_servers = DBSession.query(LuxPrintServers).all()
+        print_urls = [print_server.url for print_server in print_servers]
+        urllib2.getproxies = lambda: {}
```
rmichaelis: why this ?

rmichaelis:
LGTM

geoportailv3/views/luxprintproxy.py
geoportallux: to force urllib to not use the proxy to communicate with print although it is configured in wsgi.
until now we used the no_proxy variable, but as the hostnames are volatile, we cannot save them in the config file

PR #1911, opened 2017-11-06
[3d] Fix under terrain issue

gberaudo:
See https://jira.camptocamp.com/browse/GSLUX-16.

Switched to the Camptocamp maintained fork of Cesium.
With that version, the camera was always blocked by the terrain, as expected.


gberaudo:
Thanks for the review.

PR #1915, opened 2017-11-13
[3d] Hide and disable tools when 3d is active

tonio:
See https://jira.camptocamp.com/browse/GSLUX-6.

Based on https://github.com/camptocamp/luxembourg_geoportailv3/pull/2 , fix:
 - toggling when a incompatible tool is already open
 - toolbar appearance.

### DIFF
--- a/geoportailv3/static/js/maincontroller.js
+++ b/geoportailv3/static/js/maincontroller.js
@@ -488,6 +488,15 @@ app.MainController.prototype.createCesiumManager_ = function(cesiumURL) {
  */
 app.MainController.prototype.toggle3d = function() {
   this.ol3dm_.toggle3d();
+  // Disable uncompatible tools
+  if (!this.is3dEnabled()) { // inversed test: still not active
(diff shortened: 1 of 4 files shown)


geoportailv3/static/js/maincontroller.js
```diff
@@ -488,14 +488,23 @@ app.MainController.prototype.createCesiumManager_ = function(cesiumURL) {
  */
 app.MainController.prototype.toggle3d = function() {
   this.ol3dm_.toggle3d();
+  // Disable uncompatible tools
```
gberaudo: Typo: incompatible

gberaudo:
When the user is in "my maps" for example, I would expect the 3D button to be hidden.
Unfortunately, it is not part of the specifications and we only notice it now.

I propose we merge as-is, with this known limitation and we will discuss it during the demo. Fine with you @ybolognini?

gberaudo:
Merging it despite the typo in the code comment so that we can prepare for the demo.

PR #1916, opened 2017-11-13
[3d] widget

gberaudo:
Cf https://jira.camptocamp.com/browse/GSLUX-4.
image


geoportailv3/static/js/appmodule.js
```diff
@@ -10,16 +10,22 @@ goog.require('ngeo');
...
-app.module = angular.module('app', [ngeo.module.name, 'gettext'])
-    .run(function() {
-      if (!ol.has.TOUCH) {
-        goog.dom.classlist.add(document.body, 'no-touch');
-      }
-    });
+app.module = angular.module('app', [
+  ngeo.module.name, 'gettext',
+  ngeo.olcs.olcsModule.name,
+  ngeo.search.searchModule.module.name
```
tonio: Is that needed?

gberaudo: Yes, otherwise angular complains at runtime that it cannot find one of the search components.

PR #1921, opened 2017-11-13
Use Cesium terrain by default

gberaudo:


gberaudo:
Merging it so that it can be part ot fhe demo.

tonio:
Post-merge review: LGTM (-;

PR #1933, opened 2017-11-29
[3d] Reuse camera extent in zoom component

gberaudo:


gberaudo:
Thanks for the review.

PR #1938, opened 2017-12-04
[3d] Enable infobar in 3D

tonio:
For mouse coordinate & elevation.

### DIFF
--- a/geoportailv3/static/js/infobar/elevationdirective.js
+++ b/geoportailv3/static/js/infobar/elevationdirective.js
@@ -68,7 +68,7 @@ app.ElevationDirectiveController =
 
       // 2D
       map.on('pointermove', ngeoDebounce(function(e) {
-        if (!this['active']) {
+        if (!this['active'] || !e.coordinate) {
           return;
         }
         this.getElevation_(e.coordinate).then(
--- a/geoportailv3/static/js/infobar/elevationdirective.js
+++ b/geoportailv3/static/js/infobar/elevationdirective.js
@@ -50,10 +50,9 @@ app.module.directive('appElevation', app.elevationDirective);
  * @param {angular.$http} $http The angular http service.
  * @param {ngeo.Debounce} ngeoDebounce ngeoDebounce service.
  * @param {app.GetElevation} appGetElevation Elevation service.
- * @param {ngeo.olcs.Service} ngeoOlcsService The service.
  */
 app.ElevationDirectiveController =
-    function($scope, $http, ngeoDebounce, appGetElevation, ngeoOlcsService) {
+    function($scope, $http, ngeoDebounce, appGetElevation) {
       var map = this['map'];
 
       /**
@@ -67,24 +66,14 @@ app.ElevationDirectiveController =
        */
       this['elevation'] = '';
 
-      let manager = ngeoOlcsService.getManager();
-
       // 2D
       map.on('pointermove', ngeoDebounce(function(e) {
         if (!this['active']) {
           return;
         }
-        if (manager.is3dEnabled()) {
-          let coordinate = map.getCoordinateFromPixel(e.pixel);
-          if (!coordinate) {
-            return;
-          }
-          this['elevation'] = Math.round(coordinate[2]) + ' m';
-        } else {
-          this.getElevation_(e.coordinate).then(
-            (elevation) => (this['elevation'] = elevation)
-          );
-        }
+        this.getElevation_(e.coordinate).then(
+          (elevation) => (this['elevation'] = elevation)
+        );
       }, 300, true), this);
     };
 
--- a/geoportailv3/static/js/infobar/elevationdirective.js
+++ b/geoportailv3/static/js/infobar/elevationdirective.js
@@ -46,12 +46,14 @@ app.module.directive('appElevation', app.elevationDirective);
 /**
  * @ngInject
  * @constructor
+ * @param {angular.Scope} $scope The scope.
  * @param {angular.$http} $http The angular http service.
  * @param {ngeo.Debounce} ngeoDebounce ngeoDebounce service.
  * @param {app.GetElevation} appGetElevation Elevation service.
+ * @param {ngeo.olcs.Service} ngeoOlcsService The service.
  */
 app.ElevationDirectiveController =
-    function($http, ngeoDebounce, appGetElevation) {
+    function($scope, $http, ngeoDebounce, appGetElevation, ngeoOlcsService) {
       var map = this['map'];
 
   /**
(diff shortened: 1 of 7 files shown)


geoportailv3/static/js/infobar/projectionselectordirective.js
```diff
@@ -72,11 +81,42 @@ app.ProjectionselectorDirectiveController =
...
+          if (!this.ol3d.getEnabled()) { return; }
+          let cartesian = camera.pickEllipsoid(movement.endPosition, scene.globe.ellipsoid);
+          if (!cartesian) {
+            return;
+          }
+          let cartographic = Cesium.Cartographic.fromCartesian(cartesian);
+          cartesian = WMP.project(cartographic);
+          let coordinate = [ cartesian.x, cartesian.y ];
+          updateEl(coordinate);
+        }, 10, false), Cesium.ScreenSpaceEventType.MOUSE_MOVE);
```
gberaudo: This block is a copy paste from above.
I see two approaches:
- move the call to `getCoordinateFromPixel` in the control in an overridable method https://github.com/openlayers/openlayers/blob/v4.5.0/src/ol/control/mouseposition.js#L222; then extend the control and override the method with one aware of 3d;
- extend the ol.Map and override the existing `getCoordinateFromPixel` to make it aware of 3d.

The second solution is the most simple (and ambitious as it would fix the issue for all situations). I suggest we try this last one.

geoportailv3/static/js/infobar/elevationdirective.js
```diff
@@ -65,17 +67,46 @@ app.ElevationDirectiveController =
...
+      this['ol3d'] = null;
+
+      const requestElevation = (coordinate) =>
+        this.getElevation_(coordinate).then(
+          (elevation) => (this['elevation'] = elevation)
+        );
+
+      // 2D
+      map.on('pointermove', ngeoDebounce(function(e) {
+        if (!this['active'] || (this['ol3d'] && this['ol3d'].getEnabled())) {
```
fgravin: `ngeoOlcsService.getManager().is3dEnabled()` more consistent.

geoportailv3/static/js/infobar/elevationdirective.js
```diff
@@ -65,17 +67,46 @@ app.ElevationDirectiveController =
...
+          return;
         }
+        requestElevation(e.coordinate);
       }, 300, true), this);
+
+      // 3D
+      let unwatch = $scope.$watch(() => ngeoOlcsService.getManager().getOl3d(), (ol3d) => {
+        if (!ol3d) {
```
fgravin: Not sure `ol3d` could be null.

geoportailv3/static/js/infobar/projectionselectordirective.js
```diff
@@ -72,11 +81,42 @@ app.ProjectionselectorDirectiveController =
...
-        className: 'custom-mouse-coordinates',
-        coordinateFormat: /** @type {ol.CoordinateFormatType} */
-        (goog.bind(this.mouseCoordinateFormat_, this))
+
+      let updateEl = (coordinates) => $($element)
+        .find('.custom-mouse-coordinates')
+        .text(this.mouseCoordinateFormat_(coordinates));
+
+      this['map'].on('pointermove', ngeoDebounce((e) => {
+        if (this.old3d && this.ol3d.is3dEnabled()) {
```
fgravin: `is3dEnabled` is method of `manager`, not `ol3d`.

tonio:
Reworked with an extended map as requested.

geoportailv3/static/js/map/map.js
```diff
@@ -0,0 +1,31 @@
...
+      return coordinate;
+    }
+    const WMP = new Cesium.WebMercatorProjection(null);
+    let cartesian = manager.getCesiumScene().pickPosition(
+      new Cesium.Cartesian2(pixel[0], pixel[1])
+    );
+    if (!cartesian) {
+      return null;
+    }
+    cartesian = WMP.project(Cesium.Cartographic.fromCartesian(cartesian));
```
gberaudo: Prefer using openlayers projection transforms, passing an array created from the cartographic components.

geoportailv3/static/js/maincontroller.js
```diff
@@ -444,15 +446,14 @@ app.MainController.prototype.addLocationControl_ =
...
- * @return {!ol.Map} The map
  */
 app.MainController.prototype.createMap_ = function() {
   var interactions = ol.interaction.defaults({
     altShiftDragRotate: false,
     pinchRotate: false,
     constrainResolution: true
   });
-  var map = this['map'] = new ol.Map({
+  var map = /** @type {app.Map} */ (new app.Map({
```
gberaudo: Should not be necessary, I will look at it.

geoportailv3/static/js/infobar/elevationdirective.js
```diff
@@ -46,34 +46,44 @@ app.module.directive('appElevation', app.elevationDirective);
...
+          return;
+        }
+        if (manager.is3dEnabled()) {
+          let coordinate = map.getCoordinateFromPixel(e.pixel);
+          if (!coordinate) {
+            return;
+          }
+          this['elevation'] = Math.round(coordinate[2]) + ' m';
+        } else {
+          this.getElevation_(e.coordinate).then(
```
gberaudo: I think the luxembourg elevation is relative to the geoid. Which is not the case of Cesium.
So it is better to also retrieve the elevation through the service in all cases (to check if getElevation_ pushes or replace the 3d coordinate).

geoportailv3/static/js/infobar/projectionselectordirective.js
```diff
@@ -72,12 +81,12 @@ app.ProjectionselectorDirectiveController =
     {'label': $sce.trustAsHtml('WGS84 UTM 32|31'), 'value': 'EPSG:3263*'}
       ];
       this['projection'] = this['projectionOptions'][0];
-  /** @type {ol.control.MousePostion} */
```
gberaudo: Ideally, we should keep the original code unchanged (if possible).

geoportailv3/static/js/maincontroller.js
```diff
@@ -318,6 +319,7 @@ app.MainController = function(
   this.ol3dm_ = this.createCesiumManager_(cesiumURL, $rootScope);
   ngeoOlcsService.initialize(this.ol3dm_);
   $scope.$watch(() => this.is3dEnabled(), this.enable3dCallback_.bind(this));
+  this.map_.set('ol3dm', this.ol3dm_);
```
gberaudo: If you want you can add getter  / setter on the app.Map to store the manager.

PR #1948, opened 2017-12-18
Allow to customize the fog

gberaudo:
See https://cesiumjs.org/Cesium/Build/Documentation/Fog.html for the default Cesium values and meanings for the new URL parameters:
- fog_sse_factor
- fog_density


gberaudo:
@fgravin, these parameters are only for debugging. It is fine to keep them there.

PR #1958, opened 2018-01-17
Petzlux cms search

petzlux:
Now the search bar integrates as well the CMS page results


petzlux:
@jaykayone LGTM ?

jaykayone:
do you have a test url available?

PR #1964, opened 2018-01-29
3d mymaps: readonly mode

tonio:
Based on #1959

### DIFF
--- a/geoportailv3/static/js/maincontroller.js
+++ b/geoportailv3/static/js/maincontroller.js
@@ -407,7 +407,9 @@ app.MainController.prototype.enable3dCallback_ = function(active) {
   if (!active) {
     return;
   }
-  this['mymapsOpen'] = false;
+  if (!goog.isDef(this.ngeoLocation_.getParam('map_id'))) {
+    this['mymapsOpen'] = false;
+  }
   this['drawOpen'] = false;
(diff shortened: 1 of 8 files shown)


geoportailv3/static/js/maincontroller.js
```diff
@@ -407,7 +407,9 @@ app.MainController.prototype.enable3dCallback_ = function(active) {
   if (!active) {
     return;
   }
-  this['mymapsOpen'] = false;
+  if (!goog.isDef(this.ngeoLocation_.getParam('map_id'))) {
```
gberaudo: I think there exists an `hasParam` method in ngeo.
When the param exists it means we are not in edition?

geoportailv3/static/js/mymaps/mymapsdirective.js
```diff
@@ -135,6 +135,12 @@ app.MymapsDirectiveController = function($scope, $compile, $sce,
    */
   this.map_ = this['map'];
 
+  /**
+   * @const {?app.olcs.Lux3DManager}
+   * @export
```
gberaudo: Do you need it exported? If not it should be private.

tonio:
There are a few case that I can't test on my instance, merging now to be able to test them on dev instance.

gberaudo:
OK

PR #1967, opened 2018-01-29
Import WMS

fgravin:
- Link last version of Cesium that has a better support of custom projection.
- Set in lux layers if they support Cesium projection (4326 or 3857) via `olcs.projection` source layer param.
- Call all getCapabilities with VERSION 1.3.0 to get layer available CRS
- Casade CRS definition in getCapabilities

### DIFF
--- a/geoportailv3/static/js/wmsservice.js
+++ b/geoportailv3/static/js/wmsservice.js
@@ -114,7 +114,7 @@ app.WmsHelper.prototype.getCapabilities = function(wms) {
         basicWmsUrl = wms;
       }
       this.buildChildLayers_(basicWmsUrl, capabilities['Capability']['Layer'],
-        capabilities['version'], formats, useTiles);
+        capabilities['version'], formats, useTiles, null);
 
       return capabilities;
     }.bind(this), function(e) {
@@ -162,6 +162,7 @@ app.WmsHelper.prototype.getOnlineResource_ = function(capability, service) {
  * @return {Object} Returns the layer object.
  * @param {string} imageFormats The available formats.
  * @param {boolean} useTiles Set if the layer has a max size.
+ * @param {Object} parentLayer Parent layer object in capabilities tree.
  * @private
  */
 app.WmsHelper.prototype.buildChildLayers_ = function(wms, layer, wmsVersion,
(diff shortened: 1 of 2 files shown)


geoportailv3/static/js/wmsservice.js
```diff
@@ -181,10 +181,16 @@ app.WmsHelper.prototype.buildChildLayers_ = function(wms, layer, wmsVersion,
     layer['useTiles'] = useTiles;
   }
 
+  // casacading CRS
+  layer.CRS = layer.CRS || [];
+  if(parentLayer) {
+    parentLayer.CRS.forEach(crs => layer.CRS.indexOf(crs) < 0 && layer.CRS.push(crs));
```
tonio: this code isn't transpiled, right? So you can't use arrow functions because of [IE11](https://caniuse.com/#feat=arrow-functions)

geoportailv3/static/js/wmsservice.js
```diff
@@ -165,7 +165,7 @@ app.WmsHelper.prototype.getOnlineResource_ = function(capability, service) {
  * @private
  */
 app.WmsHelper.prototype.buildChildLayers_ = function(wms, layer, wmsVersion,
-    imageFormats, useTiles) {
+    imageFormats, useTiles, parentLayer) {
```
tonio: Doc needs to be updated

geoportailv3/static/js/wmsservice.js
```diff
@@ -468,6 +483,11 @@ app.WmsHelper.prototype.createWmsLayers = function(map, layer) {
       source: new ol.source.ImageWMS(imgOptions)
     });
   }
+  if (has3857) {
+    newLayer.getSource().set('olcs.projection', ol.proj.get('EPSG:3857'));
+  } else if (hasWGS84) {
+    newLayer.getSource().set('olcs.projection', ol.proj.get('EPSG:4326'));
```
gberaudo: Please check first if 4326 is supported.

package.json
```diff
@@ -42,9 +43,8 @@
     "less-plugin-clean-css": "1.5.1",
     "ngeo": "https://api.github.com/repos/camptocamp/ngeo/tarball/1680664d0",
     "nomnom": "1.8.1",
+    "ol-cesium": "https://api.github.com/repos/openlayers/ol-cesium/tarball/v1.34",
```
gberaudo: It should be `"ol-cesium": "1.34"`.

geoportailv3/static/js/wmsservice.js
fgravin: by what ?

gberaudo: OK, I see in Jira it is a request from the client.

PR #1977, opened 2018-02-06
[3d] Fix vector layer being detached

tonio:
Needs to be merged to check on dev instance.


gberaudo:
@tonio, what is the related Jira issue? Why do you need to merge to dev?

gberaudo:
If we need iterating before the PR is ready, I think we should consider using the 3d-demo environement.

tonio:
Looks like there isn't any related issue in Jira.
It needs to be merged because I can't manage to be logged on my dev instance, and it fix opening mymaps from authenticated user mymaps list.
Concerning `3d-demo`, I thought it shouldn't be touched between sprints.

gberaudo:
For me it is worst to merge unfinished work in 3d than to hijack the 3d-branch.
Of course we should check first with @jaykayone or @petzlux if they are OK with that and if they have finished their testings of the previous sprint release.

jaykayone:
yes we have finished testing

tonio:
@gberaudo successfully tested on `3d-demo`

PR #1982, opened 2018-02-07
rollback of change that had caused the problems with the fontawesome path

jaykayone:


jaykayone:
@gberaudo   you introduced this change several months ago.
https://github.com/Geoportail-Luxembourg/geoportailv3/commit/0582b198c1699f6e35627fa9e3e61bf478f7f649
Do you remember why? Can you check if this is OK for you to roll back

gberaudo:
Yes it is perfect. Thanks a lot for investiging the issue and fixing it.

PR #1988, opened 2018-02-13
Use stacked ground corridors

gberaudo:
See https://jira.camptocamp.com/browse/GSLUX-40.


gberaudo:
Thanks for the review @fgravin.

PR #1996, opened 2018-03-02
Fallback mymaps role when ldap is not configured

rmichaelis:
@jaykayone Are you able to check this branch in your docker environment  ?


jaykayone:
works great thanks

PR #2032, opened 2018-05-14
Add offline manifest

gberaudo:
See https://jira.camptocamp.com/browse/GSLUX-69.


gberaudo:
Thanks for the review.

PR #2035, opened 2018-05-17
Fix the sun position so that illumination is always nice

gberaudo:
See https://jira.camptocamp.com/browse/GSLUX-62.


ger-benjamin:
( And I'm in the project now :+1: )

PR #2038, opened 2018-05-24
Offline integration

fgravin:
Integration ngeo offline component in luxembourg.

- [x] place of the component map or mainCtrl ?
- [x] the way to set the template url

### DIFF
--- a/geoportailv3/static/less/offline.less
+++ b/geoportailv3/static/less/offline.less
@@ -1,28 +1,32 @@
+
 ngeo-offline {
   div {
     z-index: 1;
   }
   .main-button {
     position: absolute;
-    right: 1rem;
+    right: 8px;
     bottom: 11rem;
     cursor: pointer;
     .no-data {
       color: black;
     }
-    .with-data {
-      color: red;
-    }
     .no-data, .with-data {
       background-color: white;
       text-align: center;
-      font-size: 2.5rem;
-      line-height: 2rem;
-      border-radius: 2rem;
+      font-size: 2.7rem;
       font-family: FontAwesome;
+      line-height: 1.5em;
+      color: #2980b9;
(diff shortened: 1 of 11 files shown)


geoportailv3/static/less/offline.less
```diff
@@ -0,0 +1,58 @@
+ngeo-offline {
```
gberaudo: Have you considered using the less file from ngeo instead of duplicating it?

geoportailv3/templates/index.html
```diff
@@ -397,6 +404,7 @@ <h2 translate>Feedback</h2>
          appModule.value('ngeoLayertreeTemplateUrl', 'catalog/layertree.html');
          appModule.value('ngeoPopupTemplateUrl', 'layerinfo/popup.html');
          appModule.value('ngeoScaleselectorTemplateUrl', 'infobar/scaleselector.html');
+         appModule.value('ngeoOfflineTemplateUrl', function() {return '../node_modules/ngeo/src/offline/component.html'});
```
gberaudo: The path should be handled by pyramid. See how we did it in the 3D branch: https://github.com/Geoportail-Luxembourg/geoportailv3/blob/3d/geoportailv3/templates/index.html#L405.

geoportailv3/static/less/offline.less
fgravin: It's not ngeo core css, it's the css from the example, they are not exported in ngeo npm package.

fgravin:
@gberaudo ready for final review

fgravin:
I'll merge then.
I think we'll have to rework on CSS for final integration, for instance, it does not handle themes, but as spec are not clear for now, it's worthless to spend too much time on it.

PR #2047, opened 2018-06-07
Update ngeo dependency

gberaudo:


gberaudo:
@jaykayone, this builds locally. We will see if you still have the pipeline error when it is merged.
I am waiting for a second review for https://github.com/camptocamp/ngeo/pull/3946 and then I will merge.

PR #2055, opened 2018-06-14
Restore mymaps

ger-benjamin:
GSLUX-82

Tested via the code in the temporarily PR: https://github.com/camptocamp/ngeo/pull/3968 And Now I can restore the mymaps as expected :-)


ger-benjamin:
The `offlineRestorer.js` service is very very light for now, But I think that we will need it, as we use the `offlineDownloader.js`.

PR #2061, opened 2018-06-20
Fix GCC build

gberaudo:


geoportailv3/externs/localforage.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @return {Promise<Array<string>>}
+ */
+Localforage.prototype.keys = function() {};
+
+/**
+ * @return {Promise<number>}
+ */
+Localforage.prototype.length = function() {};
```
ger-benjamin: The same, why it doens't use the one from ngeo

gberaudo: Well, there is no such extern in ngeo.
We will remove it when switching to ES6 and building with webpack.

ger-benjamin: ok

PR #2066, opened 2018-06-21
Offline fix offline panel swicth and update ngeo

ger-benjamin:
Finish GSLUX-89

### DIFF
--- a/geoportailv3/static/js/maincontroller.js
+++ b/geoportailv3/static/js/maincontroller.js
@@ -473,8 +473,10 @@ app.MainController = function(
         return this.ngeoNetworkStatus.offline;
       }, () => {
       if (this.ngeoNetworkStatus.offline) {
-        this.closeSidebar();
-        this.layersOpen = true;
+        if (this.sidebarOpen() && !this['layersOpen'] && !this['mymapsOpen']) {
+          this.closeSidebar();
+          this['layersOpen'] = true;
+        }
         this.showTab('a[href=\'#mylayers\']');
       }
     });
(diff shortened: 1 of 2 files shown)


geoportailv3/static/js/maincontroller.js
```diff
@@ -473,8 +473,10 @@ app.MainController = function(
         return this.ngeoNetworkStatus.offline;
       }, () => {
       if (this.ngeoNetworkStatus.offline) {
-        this.closeSidebar();
-        this.layersOpen = true;
+        if (!this['layersOpen'] && !this['mymapsOpen']) {
+          this.closeSidebar();
+          this['layersOpen'] = true;
```
fgravin: Are you sure we want to open the panel while offline ?

ger-benjamin:
Thanks @fgravin 
Now I doesn't open it if it is closed.

I also fix a always true if below in the code.

geoportailv3/static/js/maincontroller.js
```diff
@@ -473,8 +473,10 @@ app.MainController = function(
         return this.ngeoNetworkStatus.offline;
       }, () => {
       if (this.ngeoNetworkStatus.offline) {
-        this.closeSidebar();
-        this.layersOpen = true;
+        if (this.sidebarOpen() && !this['layersOpen'] && !this['mymapsOpen']) {
+          this.closeSidebar();
+          this['layersOpen'] = true;
```
fgravin: Do we really favour `this['layersOpen']` over `this.layersOpen` ?

ger-benjamin: That always looks like this below in the code, so I though there is a reason for. I didn't try with the dot notation

fgravin: Ok i guess it's the way to do the `@export` thing.
Please merge.

ger-benjamin: That's not nice, but that's coherent... :-/

PR #2069, opened 2018-06-25
Merge master

ger-benjamin:
`git merge origin/master`

I had some conflict, then some problems with the `closure-util` `1.2.7` version that have a wrong / problematic `acorn` version. So now I use the last commit of  `closure-util`

Now everything looks good


package.json
```diff
@@ -45,7 +45,7 @@
     "localforage": "1.7.1",
     "ngeo": "https://api.github.com/repos/camptocamp/ngeo/tarball/f3e26bf",
     "nomnom": "1.8.1",
-    "ol-cesium": "1.34.0",
+    "ol-cesium": "https://api.github.com/repos/openlayers/ol-cesium/tarball/9bc93a85a2cf25fbec11905597f3c5cbc18acea5",
```
gberaudo: Why not take v1.37.0?

ger-benjamin: I've changed this line in a separate commit, because that was a conflict resolution error I've made. This line come simply from the master.

PR #2070, opened 2018-06-26
Copy ngeo offline component html here - caching purpose

ger-benjamin:
GSLUX-88

I've copied the file, that's the only template in this case. This solution should be enough, the template is now in `.build/build.js`

### DIFF
--- a/geoportailv3/templates/index.html
+++ b/geoportailv3/templates/index.html
@@ -438,8 +438,7 @@
          appModule.value('ngeoPopupTemplateUrl', 'layerinfo/popup.html');
          appModule.value('ngeoScaleselectorTemplateUrl', 'infobar/scaleselector.html');
          appModule.value('ngeoOfflineTestUrl', 'infobar/scaleselector.html');
-         appModule.value('ngeoOfflineTemplateUrl', function() {
-           return '${request.static_url('geoportailv3:static/js/offlineNgeoComponent.html')}'});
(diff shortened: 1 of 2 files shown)


geoportailv3/templates/index.html
```diff
@@ -439,7 +439,7 @@ <h2 translate>Feedback</h2>
          appModule.value('ngeoScaleselectorTemplateUrl', 'infobar/scaleselector.html');
          appModule.value('ngeoOfflineTestUrl', 'infobar/scaleselector.html');
          appModule.value('ngeoOfflineTemplateUrl', function() {
-           return '${request.static_url('%s/ngeo/src/offline/component.html' % node_modules_path)}'});
+           return '${request.static_url('geoportailv3:static/js/offlineNgeoComponent.html')}'});
```
gberaudo: You should not construct a static_url path but instead simply use the `offlineNgeoComponent.html` string.

gberaudo:
Regarding GSLUX-88, normally you should wait it is merged before moving it to `internal review`.

ger-benjamin:
For both comment: Oh, ok, I'll change that

geoportailv3/templates/index.html
gberaudo: (that way it will use the template cache). You should also check everything is working as expected by using the browser offline mode. You will need to re-enable the appcache for local dev to be able to test.

ger-benjamin:
I don't know why I don't have my loop problem with your string.
but anyways, I've removed the `skip_appcache = true` and it works well.

PR #2079, opened 2018-07-03
Offline state integration

fgravin:
See https://jira.camptocamp.com/browse/GSLUX-99

Set a green color to the btn when user load data from offline.


geoportailv3/static/js/offline/state.js
```diff
@@ -28,13 +28,21 @@ const OfflineState = function(ngeoNetworkStatus) {
...
+ * Is the application considered as offline.
  * @export
  */
 OfflineState.prototype.isOffline = function() {
   return this.ngeoNetworkStatus.offline || this.offline;
 };
 
+/**
+ * Get if offline mode is enabled.
+ * @export
```
gberaudo: Need the type.

geoportailv3/templates/index.html
```diff
@@ -46,7 +46,12 @@
           </div>
         </div>
 
-        <div class="offline-msg alert-danger" translate>You are currently offline.</div>
+        <div class="offline-msg alert-danger" translate>
+          <span ng-if="!mainCtrl.offlineMode.isEnabled()">You are currently disconnect.
```
gberaudo: Should be worded `You are currently disconnected.`.
In addition, you should test the network status, not the offline mode status.

geoportailv3/templates/index.html
```diff
@@ -46,7 +46,12 @@
...
+            <a ng-click="mainCtrl.offlineMode.activateOfflineMode()" translate>Enable offline mode</a>
```
gberaudo: The link should only be visible if we are not in offlineMode.

geoportailv3/templates/index.html
```diff
@@ -36,7 +36,7 @@
     <link rel="stylesheet" href="${request.static_url('geoportailv3:static/build/build.min.css')}" type="text/css">
     % endif
  </head>
-  <body data-theme="{{mainCtrl.getEncodedCurrentTheme()}}" ng-class="{offline: mainCtrl.ngeoNetworkStatus.offline, 'search-mobile': mainCtrl.mobileSearchActive, 'enabled-3d': mainCtrl.is3dEnabled()}">
+  <body data-theme="{{mainCtrl.getEncodedCurrentTheme()}}" ng-class="{offline: mainCtrl.isOfflineState(), 'search-mobile': mainCtrl.mobileSearchActive, 'enabled-3d': mainCtrl.is3dEnabled()}">
```
gberaudo: The class should be clarified:
s/offline/offline-or-disconnected/

geoportailv3/static/js/maincontroller.js
```diff
@@ -920,4 +925,13 @@ app.MainController.prototype.toggleTiles3dVisibility = function() {
   }
 };
 
+/**
+ * Check if disconnected or offline mode enabled.
+ * @returns {boolean}
+ * @export
+ */
+app.MainController.prototype.isOfflineState = function() {
```
gberaudo: The method name should be more explicit: `isOfflineOrDisconnected`.

geoportailv3/static/js/maincontroller.js
```diff
@@ -920,4 +925,13 @@ app.MainController.prototype.toggleTiles3dVisibility = function() {
...
+  return this.offlineMode.isEnabled() || !!this.networkStatus_.offline;
```
gberaudo: In ngeo, the networkStatus could be clarified and use a getter: `this.networkStatus_.isDisconnected()`.

gberaudo:
@fgravin, you will also need to rebase on #2083, which I will merge as soon as it is reviewed.

geoportailv3/templates/index.html
fgravin: Well a message is displayed if we are offline or disconnected, so there is already a test uppon in the html tree (applied with the css).
Testing here if we are not offline (like in the PR) means we are here because it is disconnected.
Otherwise i should have a double test : disconnected AND not offline mode.

geoportailv3/templates/index.html
fgravin: The test is done on the `span`

gberaudo: Said differently:
- when network is disconnected: "You are currently disconnected" should be visible;
- when not offline: "Enable offline mode" should be visible.

Your code hides the network status when offline is enabled.

PR #2083, opened 2018-07-09
Regularly store tiles and estimate tiles download size

gberaudo:


gberaudo:
The changes are minor (essentially the removal of the offline Configuration interface). Merging so that @ybolognini can start testing.

PR #2085, opened 2018-07-11
Fix appcache

gberaudo:
See https://jira.camptocamp.com/browse/GSLUX-114.


gberaudo:
Thanks for the review @fgravin.

PR #2097, opened 2018-07-18
Better calculation of size to download

ger-benjamin:
GSLUX-116

Need and wait on: https://github.com/camptocamp/ngeo/pull/4040

Now if you have a mymaps 5MB is added to the size to download.
And the "blank" bg layer count as 0MB.

So if you have the "blank" bg layer, no layer and no mymap, the size is 0MB and ngeo display a popup that prevent you to download anything.

### DIFF
--- a/geoportailv3/static/js/offline/Configuration.js
+++ b/geoportailv3/static/js/offline/Configuration.js
@@ -12,10 +12,11 @@ exports = class extends NgeoConfiguration {
   /**
    * @ngInject
    * @param {!angular.Scope} $rootScope The rootScope provider.
+   * @param {angular.$injector} $injector Main injector.
    * @param {ngeo.map.BackgroundLayerMgr} ngeoBackgroundLayerMgr Background layer
    *     manager.
    */
-  constructor($rootScope, ngeoBackgroundLayerMgr) {
+  constructor($rootScope, $injector, ngeoBackgroundLayerMgr) {
     super($rootScope, ngeoBackgroundLayerMgr);
 
     /**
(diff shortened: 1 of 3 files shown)


geoportailv3/static/js/offline/Configuration.js
```diff
@@ -37,9 +49,16 @@ exports = class extends NgeoConfiguration {
...
       if (!(layer instanceof ol.layer.Vector)) {
-        estimation += this.isBgLayer_(layer, map) ? sizeOfABgLayer : sizeOfANormalLayer;
+        if (this.isBgLayer_(layer, map)) {
+          estimation += (layer.get('label') === 'blank') ? 0 : sizeOfABgLayer;
+        } else {
+          estimation += sizeOfANormalLayer;
+        }
       }
     });
+    if (this.appMymaps_ && this.appMymaps_.getMapId() !== '') {
```
gberaudo: With mymaps, content is already downloaded. So it should add nothing, no?

geoportailv3/static/js/offline/Configuration.js
```diff
@@ -12,20 +12,32 @@ exports = class extends NgeoConfiguration {
...
     super($rootScope, ngeoBackgroundLayerMgr);
 
     /**
      * @type {ngeo.map.BackgroundLayerMgr}
      * @private
      */
     this.backgroundLayerMgr_ = ngeoBackgroundLayerMgr;
+
+    let appMymaps = null;
+    if ($injector.has('appMymaps')) {
```
gberaudo: Is it necessary to inject `appMymaps` this way?

geoportailv3/static/js/offline/Configuration.js
ger-benjamin: Ok.... that's true. In fact, I've understood (~never really read) the warning as "xxMB" will be `downloaded/stored`...
Do you think we need another method to know if there is any layer on the map ?
Or do you want that any weight is added for `mymaps` ?

geoportailv3/static/js/offline/Configuration.js
ger-benjamin: I've some errors (`"app" never provided`) if I inject `app.mymaps` directly. Maybe because we use `goog.module` in this file and `goog.provide` in mymaps ?

gberaudo: Yes, maybe. OK for the workaround.

PR #2108, opened 2018-08-02
Increase gutter to 96px and fix activation zoom to extent

gberaudo:


geoportailv3/static/js/offline/Configuration.js
```diff
@@ -3,8 +3,11 @@ goog.module.declareLegacyNamespace();
 
 goog.require('ngeo.map.BackgroundLayerMgr');
 
+const ngeoOfflineModule = goog.require('ngeo.offline.module');
 const NgeoConfiguration = goog.require('ngeo.offline.Configuration');
 
+ngeoOfflineModule.value('ngeoOfflineGutter', 96);
```
ger-benjamin: It is not already defined in ngeo ?

gberaudo: It avoid depending on a default value, which may change.
See also my last commit to allow testing different values with an URL parameter.

PR #2129, opened 2018-10-16
[WIP] Migrate to v2.3

ger-benjamin:
**DO NOT MERGE**

I do this PR now to share the state of the migration 2.3

### DB

 - Clone https://github.com/camptocamp/luxembourg_dev_db and checkout to `upgrade_to_23`
 - Take the last db dump and place it to the root of the project (`db_mapv3_30aout2018.sql`, 2.1 GB, here: http://fileserver.camptocamp.com/1_departments/1_geospatial/1_projects/3_open/luxembourg_geoportailv3/19_external_documents)
 - Go inside the project, and run `make`.

### Luxembourg

 - Clone or checkout to this branch (`upgrade_23_2n`)
 - `./docker-run --home make upgrade11`
 - `docker-compose up`

#### Run `make fix-db`

Then, you can test the application on `http://localhost:8080`.

## How to continue:
- Bugs in admin interface
  - Can't open `wms layer` tab because some wms have same id that some wmts layers. Run `SELECT * FROM geov3.layer_wms wms INNER JOIN geov3.layer_wmts wmts ON wmts.id = wms.id ORDER by wmts.layer;` to see theme. Remove matching wmts layer to see wms layers data.
  - external / internal layer bug, run: `UPDATE geov3.layergroup SET is_internal_wms = False WHERE id=807;`
- `make upgrade-db` should work and launch the 4 commands upper .
- Finish the upgrade db procedure:
  - Create a pre_GMF_23_update (on the current master) with these changes: https://github.com/camptocamp/luxembourg_dev_db/blob/upgrade_to_23/correction_for_migration.sql This must be applied **before the deploy !
- Add luxembourg new front-end.
- Test individually each function. Debug. (A lot of changes are done in the DB, and to use python 3, new configuration must be verified now).
- Clean the project ! (Identify what's deprecated in the code, and removes it)
- Retablish ldap connection (commented in geoportal/__init__.py for testing purpose)
- Build again from scratch, see if there is any error.

## Post 2.3 task
After that the production use c2cgeoportal 2.3, we have to add an alembic script to remove the column `url` of table `geov3.lux_layer_internal_wms` (now the url is in an ogc_server). We also have to remove the url in the model definition.


ger-benjamin:
Current work: Try to connect to the admin interface.
## What I know (WIP)

After the db is up to date:
- Comment this part: https://github.com/Geoportail-Luxembourg/geoportailv3/blob/upgrade_23_2nd/geoportal/geoportailv3_geoportal/__init__.py#L355-L383
- Adapt the production.ini.mako section [app] with `authtk_secure = False` and `authtkt_timeout = 86400`

Then build and run the geoportal.
Then:
 - go in the geoportal `docker exec -ti geoportailv3_geoportal_1 bash`
 - Edit `/opt/c2cgeoportal_geoportal/c2cgeoportal_geoportal/scripts/manage_users.py` with this fix: https://github.com/camptocamp/c2cgeoportal/commit/6c3e1e0df804b25c5ea8e57ff8d67e81fe835898
- Generate an "admin" user with role "role_admin":  `manage-users --create -i production.ini -r role_admin -p c2c admin` (strangely, the name to login is ['admin'], password c2c. You can rename it in the db)

Then, normally you should be able to login. but it refuses the cookie. The reason is because we are in http, and not in http. the `authtk_secure` should resolve that but it doesn't... I don't know currently why .

gberaudo:
We should use luxembourg_geoportal_1 instead of geoportailv3_geoportal_1.
It is possible to use another port than 80 for haproxy

```diff
diff --git a/Makefile b/Makefile
index 5222093..3eecc6a 100644
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ CONFIG_VARS += referrer
 CONFIG_VARS += excluded_themes_from_search
 APACHE_VHOST ?= luxembourg-geomapfish
 
-VISIBLE_WEB_HOST ?= localhost
+VISIBLE_WEB_HOST ?= localhost:8042
 
 NGEO_LIBS_JS_FILES += node_modules/fuse.js/src/fuse.min.js
 NGEO_LIBS_JS_FILES += node_modules/jszip/dist/jszip.min.js
```

Change also in docker-compose.yml `geoportal->environment->VISIBLE_WEB_HOST=localhost:8042`.


You must also set http in docker-compose.yml `geoportal->environment->VISIBLE_WEB_PROTOCOL=http`.

geoportal/geoportailv3_geoportal/scripts/finalize_c2c_23_data_adaptations.py
```diff
@@ -69,16 +69,16 @@ def main():
...
-    if treeitem is None:
-      layergroup = LayerGroup(
-        name='background',
-        is_expanded=False,
-        is_internal_wms=True,
-        is_base_layer=False
-      )
-      session.add(layergroup)
+    # Rename group bglayers to background (in GMF).
+    treeitem = session.query(TreeItem).filter(TreeItem.name == 'bglayers').one()
```
gberaudo: Should we be able to run the script several times?
If yes it would be better to continue using `one_or_none`.

ger-benjamin: Done. It's updated with one_or_none

geoportal/requirements.txt
```diff
@@ -1,3 +1,4 @@
 c2cgeoportal_commons==2.3.5.25
 c2cgeoportal_geoportal==2.3.5.25
 c2cgeoportal_admin==2.3.5.25
+cairocffi==0.9.0
```
ger-benjamin: CairoSVG doesn't work on the current version because something doens't work with python 2
There is version of CairoSVG for python 3, we should use theme instead of using a previous version.
Plus, it should be good to keep only one "requierement", see: https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/setup.py#L26 (Looks missing on our 2.3)

gberaudo: OK, done

geoportal/geoportailv3_geoportal/views/authentication.py
```diff
@@ -88,9 +88,9 @@ class O(object):
                 if 'roleOGC' in obj:
                     user.ogc_role = int(obj['roleOGC'][0])
         try:
-            user.role = DBSession.query(Role).filter_by(id=user.mymaps_role).one()
+            user.role = DBSession.query(Role).filter_by(id=roletheme).one()
         except Exception as e:
-            user.role = DBSession.query(Role).filter_by(id=default_mymaps_role).one()
```
gberaudo: @rmichaelis, why don't you define the DEFAULT_MYMAPS_ROLE environment variable?

geoportal/geoportailv3_geoportal/views/authentication.py
```diff
@@ -88,9 +88,9 @@ class O(object):
                 if 'roleOGC' in obj:
                     user.ogc_role = int(obj['roleOGC'][0])
         try:
-            user.role = DBSession.query(Role).filter_by(id=user.mymaps_role).one()
```
gberaudo: @rmichaelis, what is the meaning of `roleTheme`?

geoportal/geoportailv3_geoportal/views/authentication.py
gberaudo: In my database I only have 1 role: role_admin with id 645.
Would you be able to include the `role` table in the dump?

Having a few test maps and associated test users would be great too.
Don't you have some test users for the production db?

rmichaelis: @gberaudo The issue GSLUX-157 already contains a full export of our production database. (including role table)
Don't you work with this dump ?

geoportal/geoportailv3_geoportal/views/authentication.py
rmichaelis: @gberaudo we have several roles associated with a single user. 
"mymaps_role" associates a user  with a role dedicated to "mymaps" 
"ogc_role"  associates a user  with a role dedicated to wms stuffs.
"roletheme" associates a user with a theme.

geoportal/geoportailv3_geoportal/views/authentication.py
rmichaelis: > @rmichaelis, why don't you define the DEFAULT_MYMAPS_ROLE environment variable?

DEFAULT_MYMAPS_ROLE is dedicated to mymaps and not to themes. I could use an env variable such as DEFAULT_ROLE_THEME instead.

gberaudo: @rmichaelis, I will fix our luxembourg-dev-db values so that role_admin has id 0.
So the change LGTM.

rmichaelis: @gberaudo Or assign the role 645 to your test user.

gberaudo: Hmm, I see, I have both a role and a geov3.role table. I will check what happens.

gberaudo:
We are using a different branch so I am closing this PR.

PR #2169, opened 2019-02-14
Multi mymaps offline

ger-benjamin:
For GSLUX-175

Ongoing. Todo list in the jira ticket.

### DIFF
--- a/geoportailv3/static/js/mymapsserviceoffline.js
+++ b/geoportailv3/static/js/mymapsserviceoffline.js
@@ -42,13 +42,13 @@ app.MymapsOffline = function(appMymaps, appDrawnFeatures, ngeoOfflineConfigurati
  * Save data into the storage system.
  */
 app.MymapsOffline.prototype.save = function() {
-  var item = {
-    'allCategories': this.appMymaps_.allcategories,
-    'mapInfo': this.appMymaps_.getMapInfo(),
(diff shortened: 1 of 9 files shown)


geoportailv3/static/js/mymapsserviceoffline.js
```diff
@@ -82,4 +83,20 @@ app.MymapsOffline.prototype.restore = function() {
...
+ * Check if the stored data has the new format (multi-mymaps)
+ * If no, clear the cache
+ */
+app.MymapsOffline.prototype.checkDataFormat = function() {
+  this.ngeoOfflineConfiguration_.getItem(this.storageGroupeKey_).then((storedItem) => {
+    if (!storedItem) {
+      return;
+    }
+
+    if (!storedItem['data_version']) {
```
gberaudo: You should test the exact number (have it as a file constant).

geoportailv3/static/js/maincontroller.js
```diff
@@ -536,6 +544,8 @@ app.MainController = function(
         this['layersOpen'] = true;
       }
       this.showTab('a[href=\'#mylayers\']');
+    } else {
+      this.appMymapsOffline_.checkDataFormat();
```
gberaudo: This could be done in the mymapsserviceoffline constructor.
We don't really care if the user is offline or not.

PR #2331, opened 2019-10-10
Replace raster layer by vector one

gberaudo:
See https://jira.camptocamp.com/browse/GSLUX-262.


jaykayone:
attention, this should also work for the case where we are in retina. the name od the layer to replace is different in that case

jaykayone:
permalinks to the old layer should also still work

PR #2505, opened 2020-06-12
Render measure in print.

tonio:
Transition from HTML tooltips to ol styles.

The sketchStyle class ngeo methods to get measure, store it in feature property on `measureend`, & final style use it, because text in tooltip is not available anymore.
Feature is stored & readded to layer when active tool changes.

A `hidden` property is added to measure layer, which prevent it to be displayed in layer manager.

Needs [small ngeo tweak](https://github.com/camptocamp/ngeo/commit/b178a9e7451a4b660c728fedcf015e9f9854cd50).

(last minute check show me azimut isn't drawn in print… is `Circle` supported in MF print?)

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
@@ -799,6 +799,13 @@ const MainController = function(
   this['selectedLayers'] = [];
 
   /**
+   * @type {function}
+   */
+  this['selectedLayersLength'] = function() {
+    return this['selectedLayers'].filter(l => !l.get('metadata').hidden).length
+  }
+
+  /**
    * @type {Array}
    */
   this['ageLayers'] = [];
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
@@ -801,8 +801,8 @@ const MainController = function(
   /**
    * @type {function}
    */
-  this['selectedLayersLength'] = function() {
-    return this['selectedLayers'].filter(l => !l.get('metadata').hidden).length
+  this.selectedLayersLength = function() {
+    return this.selectedLayers.filter(l => !l.get('metadata').hidden).length
   }
 
   /**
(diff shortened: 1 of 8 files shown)


gberaudo:
@tonio, I would say circles are supposed to work. They are mentioned in the doc: https://mapfish.github.io/mapfish-print-doc/styles.html#stylev2

gberaudo:
Reset does not seem to work for lenghtes.

geoportal/geoportailv3_geoportal/static-ngeo/js/measure/MeasureController.js
```diff
@@ -104,14 +144,20 @@ const exports = function($scope, $q, $http, $compile, gettext,
       fill: new olStyleFill({
         color: 'rgba(255, 204, 51, 0.3)'
       })
+    }),
+    text: new olStyleText({
+      font: 'bold 12px Calibri,sans-serif',
+      fill: new olStyleFill({
+        color: 'rgb(0 0 0 / 100)'
```
gberaudo, line 183: I don't know this notation, what does it do?

geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
```diff
@@ -798,6 +798,13 @@ const MainController = function(
    */
   this['selectedLayers'] = [];
 
+  /**
+   * @type {function}
+   */
+  this['selectedLayersLength'] = function() {
```
gberaudo, line 804: Brackets were used for closure-compiler. It is not necessary anymore. Better not to use them in new code.

geoportal/geoportailv3_geoportal/static-ngeo/js/measure/MeasureController.js
tonio: Should have been `rgb(0 0 0 / 100%)` (or even w/o the `/ 100%`, the default), see https://mobile.twitter.com/mathias/status/1253242715304857601 for ref.

geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
tonio: I wondered but keep brackets for coherence with the rest of the file, I can change if needed.

tonio:
@gberaudo this one is ready from my side. Is it good to merge for you?

PR #2538, opened 2020-09-01
Tweak CSS to remove the scroll due to footer override

RemiDesgrange:
Before: 
image
Now
image

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/less/layout.scss
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/less/layout.scss
@@ -52,7 +52,7 @@ app-search {
 
 .footer {
   position: absolute;
-  bottom: 2px;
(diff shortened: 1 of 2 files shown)


geoportal/geoportailv3_geoportal/static-ngeo/less/layout.scss
```diff
@@ -52,7 +52,7 @@ app-search {
 
 .footer {
   position: absolute;
-  bottom: 0;
+  bottom: 2px;
```
tonio, line 55: Looks like a magic value that need some explanations…

RemiDesgrange: That's the only value (and way) to make it work. Sounds like a hack..

tonio (approved):

PR #2539, opened 2020-09-01
Changement style 3d

llienher:
https://jira.camptocamp.com/browse/GSLUX-353


tonio (approved):
Code looks good, half-tested locally, seems to do the job! thx

PR #2541, opened 2020-09-02
Fix selected simple styles

llienher:
https://jira.camptocamp.com/browse/GSLUX-366


tonio (approved):
Tested, works fine, thanks!

PR #2542, opened 2020-09-03
Avoid extra blank page in landscape layout

tonio:
GSLUX-350


llienher (approved):
Tested with landscape A4 (pdf),  works for me !

PR #2543, opened 2020-09-04
Error print in console

llienher:
https://jira.camptocamp.com/browse/GSLUX-372

First commit is just a little cleanup, does not fix anything.

To test, just open the print panel with a serialized config in the permalink and a non-vector tile active background.


tonio (approved):
:+1:

PR #2547, opened 2020-09-08
Esri legends

mki-c2c:
enable pulling legends from arcgis server including localisation and DPI awareness


geoportal/geoportailv3_geoportal/views/legends.py
```diff
@@ -41,15 +50,70 @@ def get_png(self):
...
+                        html_legend += '<table class="inline">\n'
+                        for leg in l['legend']:
+                            html_legend += '<tr class="row0"><td class="col0">\n'
+                            html_legend += '<img alt="" class="media" src="data:image/png;base64,%s" style="max-width:32px;"/>\n' % leg['imageData']
+                            html_legend += '</td><td class="col1" style="padding-left:5px">%s</td></tr>\n' % localizer.translate(
+                                legend(internal_wms.layer + " / " + l['layerName']
+                                       + " : " + leg['label'])
+                            )
+
+                        html_legend += '</table></div></div>'
```
tonio, line 103: I'd have used a separate template file for readability.

tonio (commented):
Good job! Mainly small comments atm.

geoportal/geoportailv3_geoportal/lib/esri_authentication.py
```diff
@@ -0,0 +1,117 @@
...
+
+ResultTuple = namedtuple('ResultTuple', ['data', 'content_type'])
+
+
+def read_request_with_token(url_request, parent_request, log, timeout=15, renew_token=True):
+    result = urllib.request.urlopen(url_request, None, timeout)
+    data = result.read()
+    try:
+        resp = json.loads(data)
+    except Exception as e:
```
RemiDesgrange, line 21: Maybe use a less generic exception handler ? (JSONDecodeError in this case)

geoportal/geoportailv3_geoportal/lib/esri_authentication.py
```diff
@@ -0,0 +1,117 @@
...
+                url = urllib.parse.urlunsplit(url_tuple)
+                url_request.full_url = url
+
+                # Try to re-read with new token
+                result = urllib.request.urlopen(url_request, None, timeout)
+                data = result.read()
+
+            try:
+                resp = json.loads(data)
+            except Exception as e:
```
RemiDesgrange, line 44: Same as above.

geoportal/geoportailv3_geoportal/lib/esri_authentication.py
```diff
@@ -0,0 +1,117 @@
...
+import requests
+import urllib
+from collections import namedtuple
+
+
+class ESRITokenException(Exception):
+    pass
+
+
+ResultTuple = namedtuple('ResultTuple', ['data', 'content_type'])
```
RemiDesgrange, line 13: Since we are in python 3.8 (or 3.7). You could use a dataclass.

geoportal/geoportailv3_geoportal/lib/lingua_extractor.py
```diff
@@ -0,0 +1,378 @@
...
+        try:
+            engine = sqlalchemy.engine_from_config(self.config, "sqlalchemy_slave.")
+            factory = sqlalchemy.orm.sessionmaker(bind=engine)
+            db_session = sqlalchemy.orm.scoped_session(factory)
+            c2cgeoportal_commons.models.DBSession = db_session
+            c2cgeoportal_commons.models.Base.metadata.bind = engine
+            for dbsession_name, dbsession_config in self.config.get("dbsessions", {}).items():  # pragma: nocover
+                c2cgeoportal_commons.models.DBSessions[dbsession_name] = c2cwsgiutils.db.create_session(
+                    config, dbsession_name, **dbsession_config
+                )
```
RemiDesgrange, line 61: I'm more in favor of decoupling responsibility. This class should take the ORM session and the config as argument to the class.

geoportal/geoportailv3_geoportal/lib/lingua_extractor.py
```diff
@@ -0,0 +1,378 @@
...
+
+        from c2cgeoportal_commons.models import DBSession  # pylint: disable=import-outside-toplevel
+        from geoportailv3_geoportal.models import LuxLayerInternalWMS
+
+        results = (DBSession.query(LuxLayerInternalWMS)
+                   .filter(LuxLayerInternalWMS.rest_url.isnot(sqlalchemy.null)))
+        print("%d ESRI layers to parse" % results.count())
+
+        for result in results:
+            self._load_result(result)
```
RemiDesgrange, line 143: `dbsession` is a already a parameter of the fonction, why importing it ?

Also we are using a recent version of python, you can drop the old style print and do `print(f"{results.count()} ESRI layers to parse")` But I'm maybe to strict :sweat_smile:

geoportal/geoportailv3_geoportal/lib/lingua_extractor.py
```diff
@@ -0,0 +1,378 @@
...
+        except Exception as e:
+            log.exception(e)
+            return []
+        return []
+
+    def _extract_messages(self, session):  # pragma: nocover
+        print('Entering into the tooltip extractor')
+
+        from c2cgeoportal_commons.models import DBSession, DBSessions
+        from geoportailv3_geoportal.models import LuxGetfeatureDefinition, LuxLayerInternalWMS
```
RemiDesgrange, line 314: Same, Database session seams like an arg of the function.

geoportal/geoportailv3_geoportal/lib/lingua_extractor.py
```diff
@@ -0,0 +1,378 @@
...
+        print("%d results" % results.count())
+        for result in results:
+            engine = DBSessions[result.engine_gfi]
+            first_row = None
+            if result.query is not None and len(result.query) > 0:
+                try:
+                    if "SELECT" in result.query.upper():
+                        first_row = engine.execute(result.query).first()
+                    else:
+                        first_row = engine.execute("SELECT * FROM " + result.query).first()
```
RemiDesgrange, line 331: This my friend, is SQL injection, via SQL :exploding_head: 

When using raw SQL, I tend to wrap everything with the `text()` function of SQLAlchemy.

RemiDesgrange (changes requested):
All in all it represents a huge amount of work (congrat @mki-c2c !)  But without tests that seems pretty fragile since we are integrating a third party API which could change anytime.

geoportal/geoportailv3_geoportal/lib/esri_authentication.py
mki-c2c: @RemiDesgrange 
I remember I tried this, but the JSON Error does not work.
In this case I get a UnicodeDecodeError, but I think it might also be a JSONDecodeError in some cases, so I'm not quite sure if I can use a less generic Exception.
Any Ideas ?

geoportal/geoportailv3_geoportal/lib/esri_authentication.py
mki-c2c: not worth the effort, IMO
And not compatible with current use unpack tuple AND named access

geoportal/geoportailv3_geoportal/lib/lingua_extractor.py
mki-c2c: yes, this was a bit of a mess -> cleanup

geoportal/geoportailv3_geoportal/lib/lingua_extractor.py
mki-c2c: yes

geoportal/geoportailv3_geoportal/lib/lingua_extractor.py
mki-c2c: done

geoportal/geoportailv3_geoportal/lib/lingua_extractor.py
mki-c2c: this is a legacy configuration, the configuration DB can contain raw SQL queries. (it is controlled by an admin though)
Hard to refactor
we should discuss this

RemiDesgrange: At least using the `text()` function would mitigate a lot of cases.

PR #2550, opened 2020-09-14
Serial is not initialized at launch if existing in local storage

llienher:
https://jira.camptocamp.com/browse/GSLUX-374

I changed the local storage behavior if there is already a expert or medium style saved. Now only the one active is saved there, the other is removed if existing. That insures that we can get only one as we launch the app without the permalink complete and can fill it with accurate local storage data without having two styles conflicting. This actually removes unused data from local storage.

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
@@ -192,25 +192,7 @@ exports.prototype.onLayerUpdate_ = function(layers) {
   var bgLabel = 'blank';
   if (bgLayer) {
     bgLabel = bgLayer.get('label');
-
-  // If there is a expert mvt style in the localstorage but no parameter in the url
-    const itemKey = 'remoteIdForStyle_' + bgLabel;
-    const stateMvtExpertStyle = this.stateManager_.getValueFromLocalStorage(itemKey);
-    if (stateMvtExpertStyle !== undefined && stateMvtExpertStyle !== null) {
-      this.ngeoLocation_.updateParams({
-        'serial': stateMvtExpertStyle
-      });
-    }
   }
-
-  // If there is a medium mvt style in the localstorage but no parameter in the url
-  const stateMvtMediumStyle = this.stateManager_.getValueFromLocalStorage('mediumStyling');
-  if (stateMvtMediumStyle !== undefined && stateMvtMediumStyle !== null) {
-    this.ngeoLocation_.updateParams({
-      'serial': stateMvtMediumStyle
-    });
-  }
-
   this.stateManager_.updateState({
     'layers': layerIds.join('-'),
     'opacities': opacities.join('-'),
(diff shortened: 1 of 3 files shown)


geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
```diff
@@ -192,7 +192,25 @@ exports.prototype.onLayerUpdate_ = function(layers) {
   var bgLabel = 'blank';
   if (bgLayer) {
     bgLabel = bgLayer.get('label');
+
+  // If there is a expert mvt style in the localstorage but no parameter in the url
+    const itemKey = 'remoteIdForStyle_' + bgLabel;
+    const stateMvtExpertStyle = this.stateManager_.getValueFromLocalStorage(itemKey);
+    if (stateMvtExpertStyle !== undefined && stateMvtExpertStyle !== null) {
```
tonio, line 199: What about just:
```suggestion
    if (stateMvtExpertStyle) {
```

geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
```diff
@@ -192,7 +192,25 @@ exports.prototype.onLayerUpdate_ = function(layers) {
   var bgLabel = 'blank';
   if (bgLayer) {
     bgLabel = bgLayer.get('label');
+
+  // If there is a expert mvt style in the localstorage but no parameter in the url
+    const itemKey = 'remoteIdForStyle_' + bgLabel;
```
tonio, line 197: maybe this pattern should not be repeated here but imported… not blocking btw.

tonio (approved):
Tested, works fine.
Code looks good, thanks!

geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
llienher: I did not imported it from MvtStylingService as I would need to import the whole service into the LayerPermalinkManager just for a single function to use... So I just copied the needed line to get the correct label here.

llienher:
@tonio I had to completely refactor my PR as I found a new bug just changing the requested lines you mentionned... :-(
That means you need to review it again.

The bug was that in case of local storage, but empty permalink, the map would override to default style as it is what it would do with empty permalink. This would not display the custom style (for mediumStyle at least, I guess expoert was loaded correctly).

I had to my code to MvtStyleService to make sure the permalink objects are set with data from the local storage before giving the style to OpenLayers at application loading.

tonio (approved):
Untested, but code looks good. Much easier to read. Thanks!

rmichaelis:
@llienher Est-ce que cette PR est prête pour être mergée ?

tonio:
@rmichaelis oui, l'erreur de CI est lié à un pb sur une dépendance mais qui sera fixé dans la branche d'upgrade. Je propose de merger cette PR afin que vous puissiez valider sur migration.geoportail.lu. Il me semble que c'était le dernier blocker pour la mise en prod.

rmichaelis:
Merci pour l'info. Tu peux merger quand tu veux.

PR #2553, opened 2020-09-22
reactivate the jsapi

RemiDesgrange:
Also remove PGOPTIONS env which breaks db connection.


tonio (approved):
I have no problems with the db connection, it may have been splitted in 2 separated commits.
We'll have to check jsapi again once backend is upgraded.

PR #2554, opened 2020-09-22
Fix rotate in map

RemiDesgrange:


tonio:
Pinch rotate doesn't look activated on https://apiv3.geoportail.lu/api-example you may have to check what's needed in API too. (maybe check ol defaults).

PR #2556, opened 2020-09-29
fix: mobile buttons unexpectedly showing on desktop app

tonio:
GSLUX-384


llienher (approved):
I did not tested, but looks good to me !

PR #2557, opened 2020-09-29
fix: avoid hidden layer to be wrongly added to permalink

tonio:
That breaks loading layers from permalink.

GSLUX-383


llienher (approved):
Tested and works correctly.

PR #2559, opened 2020-10-02
Fix style of admin page `predefined WMS`

mki-c2c:
a bug has been observed since geoforms uses by default the name of the column for the CSS class. This results in a clash for the column 'label' of the table `lux_predefined_wms`

Before
image

After
image


tonio (approved):
:+1:

RemiDesgrange (approved):

PR #2569, opened 2020-10-07
Fix permalink parameters disappearing

llienher:
GSLUX-361

Seems it was removed from ngeo 2.5 then from the project with migration by Guillaume. Just added it back to the project.

We need to check why we need anchorScroll in DrawController, as there is possibly  some issues with that as the comments in the file I copied mentionned.


tonio (approved):
Tested, works fine, thanks!

PR #2572, opened 2020-10-07
GMF 2.5 uses python 3.7 (and ubuntu 18.04).

RemiDesgrange:
To be more consistent, use the same version

### DIFF
--- a/config/Dockerfile
+++ b/config/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM python:3.7
 LABEL maintainer Camptocamp "info@camptocamp.com"
 ARG HTTP_PROXY_URL
 ENV http_proxy $HTTP_PROXY_URL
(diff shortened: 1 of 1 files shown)


config/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM python:3.7
```
mki-c2c, line 1: can you try:
FROM camptocamp/geomapfish-tools:2.5.0.77

to be consistent with the rest of the structure ?

mki-c2c (changes requested):

config/Dockerfile
mki-c2c: actually the scaffold system of gmf2.5 places the config image into . instead of ./config and builds from gmf-tools

https://github.com/camptocamp/c2cgeoportal/blob/2.5/geoportal/c2cgeoportal_geoportal/scaffolds/create/build_tmpl#L109
https://github.com/camptocamp/c2cgeoportal/blob/2.5/geoportal/c2cgeoportal_geoportal/scaffolds/create/Dockerfile_tmpl#L1

RemiDesgrange: There is no "camptocamp/geomapfish-tool:2.5". I have to pick a very specific tag like 2.5.0.80. Also it weight 2.25Go :scream:

PR #2585, opened 2020-10-16
Suppressing dead or misleading code

RemiDesgrange:
Following PyCharm recommandation and refactoring tools. renamming
misleading variable name (like id, map etc...) which are built-in
function in python.

We also should add `black` checks.


RemiDesgrange:
Migration to 2.5 solve all of this stuff

PR #2586, opened 2020-10-16
Update upstream_requirements.txt

mki-c2c:
Found a minor bug with Arnaud:
Chameleon 3.8 is incompatible with deform 2.0.7
this leads to the problem that checkboxes for boolean values are checked by default in admin interface

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ help:
 build: docker-build-geoportal docker-build-config docker-build-print docker-build-ldap
 
 .PHONY: build-prod
-build: docker-build-geoportal docker-build-config
+build-prod: docker-build-geoportal docker-build-config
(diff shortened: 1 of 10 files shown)


tonio:
I thought it was fixed with https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportal/luxembourg_requirements.txt#L24 ?

tonio:
But the need of all this different requirements files is confusing.

mki-c2c:
ah, I ran into this in a branch, so it was not up to date.

however, it would be good to have Chameleon and deform requirements in the same file (either lux or upstream)
(including a comment about the incompatibility)
so why not remove it from lux_requirements ?

PR #2593, opened 2020-10-30
Add db to composition

RemiDesgrange:


mki-c2c (approved):
Sounds good, it's the right compromise between manual build and automatic launch

tonio (approved):
:shipit:

PR #2594, opened 2020-11-04
Add working wintermesh

RemiDesgrange:
I bump the version of `@camptocamp/cesium`. It doesn't seems to work with a "plain" cesium.

It also apprear that building are not always shown when wintermesh is displayed. Data (b3dm files) are collected from the server but not displayed. Working on a fix.

To be noted, I had to add `ARCGIS` env var to get the composition to start.


geoportal/geoportailv3_geoportal/static-ngeo/js/olcs/Lux3DManager.js
```diff
@@ -131,7 +131,7 @@ const exports = class extends ngeoOlcsManager {
   init3dTiles(visible) {
     const scene = this.ol3d.getCesiumScene();
     this.tiles3dLayers_.forEach((layer) => {
-      const url = this.tiles3dUrl_ + layer;
+      const url = this.tiles3dUrl_ + layer + "/tileset.json";
```
tonio, line 134: you may have checked if `tileset.json` is present before adding it.

tonio (approved):
Code looks, just a small comment.
Performance feels degraded to me, not sure if related to the mesh or cesium upgrade.

Let's deploy to confirm or not.

geoportal/geoportailv3_geoportal/static-ngeo/js/olcs/Lux3DManager.js
RemiDesgrange: Like making an HTTP requests to the `tileset.json` ?

tonio: is present *in the url* (-:

PR #2616, opened 2020-11-26
Fix featuresearch 2

mki-c2c:
Bugfixes according to comments in https://jira.camptocamp.com/browse/GSLUX-392
Corrections for the following problems:
- fix case insensitive search
- fix bbox for empty geometries
- fix id if missing
- enable extent filter for feature search

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/search/SearchController.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/search/SearchController.js
@@ -742,6 +742,14 @@ exports.prototype.createAndInitFeatureBloodhoundEngine_ =
           let selected_layers = this.selectedLayers.map((el) => el.get('queryable_id'))
             .filter(el => el !== undefined);
           params.set('layers', encodeURIComponent(selected_layers.join(',')));
+          if (this.facets.extent) {
+            let extent = transformExtent(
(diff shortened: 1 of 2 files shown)


geoportal/geoportailv3_geoportal/static-ngeo/js/search/SearchController.js
```diff
@@ -742,6 +742,14 @@ exports.prototype.createAndInitFeatureBloodhoundEngine_ =
...
             .filter(el => el !== undefined);
           params.set('layers', encodeURIComponent(selected_layers.join(',')));
+          if (this.facets.extent) {
+            let extent = transformExtent(
+              this.map.getView().calculateExtent(),
+              'EPSG:3857',
+              'EPSG:4326'
+            );
+            params.set('extent', extent.join(','))
+          }
```
tonio, line 752: couldn't this have been shared rather than copied? (-:

tonio (approved):
LGTM, please just try to avoid code duplication (-;

geoportal/geoportailv3_geoportal/static-ngeo/js/search/SearchController.js
mki-c2c: I tried refactoring here, but the result seems less readable to me. I would prefer to leave it this way.

PR #2621, opened 2020-12-01
Fix themes.json with internal WMS

arnaud-morvan:
Fake capabilities for OGC server "Internal WMS" as capabilities is not available for this WMS proxy.
@tonio (for review)


tonio (approved):
Tested, works great.
Much easier to customize, thanks a lot!

PR #2637, opened 2020-12-11
Increase url size limit of gunicorn

rmichaelis:


llienher (approved):
Looks good to me, I tested and it's working (local in dev mode).

PR #2639, opened 2020-12-11
Apply correctly the topo config from permalink

llienher:


tonio (approved):
Untested, but looks very good, thanks.

PR #2785, opened 2021-09-22
Make the auth cookie valid forever for mobile app

fredj:
work in progress


tonio (approved):
Untested, but looks like a good way to do it. (-:

geoportal/geoportailv3_geoportal/lib/lux_authentication.py
```diff
@@ -22,16 +19,30 @@ def create_authentication(settings):
     http_only = http_only.lower() in ("true", "yes", "1")
     secure = settings.get("authtkt_secure", "True")
     secure = secure.lower() in ("true", "yes", "1")
-    cookie_authentication_policy = AppAwareAuthTktAuthenticationPolicy(
+
+    # AuthenticationPolicy for login via the app
+    app_authentication_policy = ConditionalAuthTktAuthenticationPolicy(
```
RemiDesgrange, line 24: I'll rename "app" to "mobile app" (maybe just the comment) for clarity.

fredj: I've updated the comment

PR #2870, opened 2022-03-18
jsapi improvements 

fredj:


fredj:
Merging to test deployment on https://migration.geoportail.lu/api-example

PR #2913, opened 2022-07-12
fix longpress on mobile

mki-c2c:


tkohr (commented):
Thanks @mki-c2c, code LGTM. Longpress works smoothly when testing on chromium mobile emulator, but not on firefox mobile emulator (might just be related to the emulator). I see that the `else if` clause is entered here as well, but info panel does not open. Not tested on iOS.

mki-c2c:
> Thanks @mki-c2c, code LGTM. Longpress works smoothly when testing on chromium mobile emulator, but not on firefox mobile emulator (might just be related to the emulator). I see that the `else if` clause is entered here as well, but info panel does not open. Not tested on iOS.

let's check this afterwards. the change was working on android and IOS SDK emulators

PR #2914, opened 2022-07-13
fix search results in IOS

mki-c2c:
on small screens with a big number of results, items far down cannot be reached

the results list is therefore made scrollable (auto)

however in IOS, the viewport is not reduced by the keyboard overlay, so the whole screen may become scrollable
this seems to be a hard to fix undesirable feature of IOS
 => https://selleo.com/til/posts/l8xyxi1sao-preventing-body-scroll-for-modals-in-ios
 => https://stackoverflow.com/questions/37713970/ios-css-js-overlay-scroll-but-prevent-body-scroll


tkohr (approved):
Thanks @mki-c2c, changes LGTM. Scrolling and unblocking focus work fine when testing on chromium and firefox. Not tested on iOS.

PR #2927, opened 2022-08-26
make lidarProfileUrl configurable.

rmichaelis:


MaYohann:
Makes more sens that way, thanks!

MaYohann (approved):

PR #2947, opened 2022-10-14
Update geoportail-icons font for theme icons

tkohr:
Currently, the icons for the themes in the geoportail come from four different fonts. This PR adds all Luxembourg specific icons from "Geoportail-Grande-Region", "urban-farming", "statec-icons" fonts to "geoportail-icons" and uses them for the themes. This will allow simplifying this part during the migration to web components and using one icon font in the long term.

I'm still wondering if we need to update the other formats as well or if `woff` is sufficient.

Note: Other parts of the application might still rely on the other fonts, which is why they will not (yet) be removed. The goal should be to use icons from the "geoportail-icons" font in all components step by step, when migrating the according components. One excellent tool to edit fonts is fontforge (in case more icons still need to be added in the future).

Extract of the Luxembourg icons in `geoportail-icons.woff` with accoridng codes:

lux-icons


mki-c2c:
restent à migrer:
grande region
e903
e904
e905
e907
e908
etc

mki-c2c (commented):
looks good (on chrome and edge v100.0.1185.29)

@tkohr don't you have to merge also the other font formats (.eot, .tff, .svg) ?
If I understand well, these formats are not supported in chrome and firefox, but might be loaded in edge or safari ??
https://www.w3schools.com/cssref/css3_pr_font-face_rule.php

why is woff loaded here as the second source ?
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/4bfff45f7a306f40bf75c888dab526db1387e421/geoportal/geoportailv3_geoportal/static-ngeo/less/icons.less#L15

tkohr:
Thanks for your attention @mki-c2c ! You're right, I should have adapted the CSS font definitions. Regarding the format, we figured that `woff` should be sufficient nowadays.

Looking again at the situation, I had the impression, that it would to easier to understand in the future, if the manipulated font was renamed and just provided for the new web components. The current components/icons will not even need to be modified in this case. So this is what I did, in my latest commit, hoping you agree with this approach. I will make a PR in the WC repo for the renaming.

mki-c2c (approved):

PR #2971, opened 2022-11-14
[Offline mode] Inform the user if an error occured

tkohr:
PR displays a modal to the user if an error during the vector tile downloads occurred, that is, when the local backend answers with an inconsistent response.


tkohr:
Thanks for the review and feedback @mki-c2c. I reworked the PR from zero, as we discussed earlier, and will let you have another look at it :-)

mki-c2c (approved):

PR #2973, opened 2022-11-21
Ngeo offline: Display button correctly

tkohr:
The ngeo-offline button was broken by https://github.com/Geoportail-Luxembourg/geoportailv3/commit/412f1eb0722a198d4a0aa9126f9ace5448a97f7c (which modified the button display for the native iOS app)

This PR should fix the 'arrow down' icon to be displayed in green (active) or red (inactive). The CSS `!important` is needed to overwrite the `ol-control` color, a class added in the iOS fix above.

To-do:

- [ ] Test button on iOS
- [x] Verify download process works fine => layers (other than background) are currently not displayed after download (on production and migration)


tkohr:
Following our tests together yesterday, I will merge this to fix the button display @mki-c2c. This does not touch on the ngeo offline functionality.

PR #3078, opened 2023-07-31
Gslux 637 fix dev build

mki-c2c:
remove hard source plugin for more reliable builds and split dev build to enable faster rebuilds


AlitaBernachot:
@mki-c2c Should we close this one? I think it is out of date.

PR #3081, opened 2023-08-02
GSLUX-631: Background selector KO

AlitaBernachot:
Please test this PR with this v3 PR : https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/pull/63

### JIRA issue

https://jira.camptocamp.com/browse/GSLUX-631

### Description

Background selection was not working as expected in v3.

- enable blank bg selection in v3 (bg `ZIndex` is now `-200` (instead of -1) like in v3 legacy
- handle special case: when **bg layer is blank** in permalink **and no layers** in permalink (was not working)
- added new component for alert notifications (when bg layer is blank in permalink and no layers in permalink, it now displays the message in an alert box, instead of display message with a `console.log`)
- added e2e tests

Some more stuffs:
- npm commands for building lib dev: disable minification in dev mode for debug purpose
- do not import main.css in ol.css, causing duplicated styles in the dom

### Screenshots

image

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
@@ -622,12 +622,19 @@ exports.prototype.initBgLayers_ = function() {
       stateBgLayerOpacity = this.ngeoLocation_.getParam('bgOpacity');
     }
     var hasBgLayerInUrl = (this.ngeoLocation_.getParam('bgLayer') !== undefined);
-    if (mapId === undefined || hasBgLayerInUrl) {
-      var layer = /** @type {ol.layer.Base} */ (bgLayers.find(layer => layer.get('label') === stateBgLayerLabel));
-      if (layer !== undefined) {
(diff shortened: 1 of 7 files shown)


geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
```diff
@@ -622,12 +622,12 @@ exports.prototype.initBgLayers_ = function() {
       stateBgLayerOpacity = this.ngeoLocation_.getParam('bgOpacity');
     }
     var hasBgLayerInUrl = (this.ngeoLocation_.getParam('bgLayer') !== undefined);
-    if (mapId === undefined || hasBgLayerInUrl) {
```
tkohr, line 625: Do we also handle the `mapId === undefined` case in v4? Is this related to mymaps?

AlitaBernachot: Thanks for noticing this!
I have added this use case in the default background id selection.

tkohr (approved):
Thanks a lot @AlitaBernachot ! Everything works great in the v3 integration now :-)

PR #3082, opened 2023-08-03
GSLUX-633: On click theme button in header

AlitaBernachot:
This PR needs: https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/pull/64

### JIRA issue

https://jira.camptocamp.com/browse/GSLUX-633

### Description

When clicking on theme button in the header, it should: 
- display the side panel (if not already displayed) and display the theme grid
- if the side panel is visible but the catalog tab is not selected, it should select the catalog tab and display the theme grid
- if the side panel and the theme grid are already visible, it should hide the side panel

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
@@ -1963,6 +1963,8 @@ MainController.prototype.showTab = function(selector) {
  * @export
  */
 MainController.prototype.toggleThemeSelector = function() {
+  const { layersOpen, myLayersTabOpen, themeGridOpen } = storeToRefs(useAppStore())
+
   var layerTree = $('app-catalog .themes-switcher');
   var themesSwitcher = $('app-themeswitcher #themes-content');
   var themeTab = $('#catalog');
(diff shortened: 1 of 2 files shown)


geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
```diff
@@ -1969,6 +1971,19 @@ MainController.prototype.toggleThemeSelector = function() {
...
     layerTree.collapse('hide');
   }
+
+  if (!layersOpen.value) {
+    useAppStore().setLayersOpen(true)
+    myLayersTabOpen.value && useAppStore().setMyLayersTabOpen(false)
+    useAppStore().setThemeGridOpen(true)
+  } else if (layersOpen.value) {
+    if (themeGridOpen.value) {
+      useAppStore().setLayersOpen(false)
```
tkohr, line 1981: I think when another tool in the footer bar is open and the themegrid has been opened before (and still has an open state), the code can fall into this condition and close the sidepanel instead of showing the themegrid.

tkohr (commented):
Thanks, @AlitaBernachot! Works fine besides the one case that I mentioned in the code. To reproduce: Open theme grid => Open another tool in the footer bar => Click on theme button in header

geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
AlitaBernachot: yes, thanks for reporting this, I have made a fix.

tkohr: Thanks for the fix, works as expected now :-)

tkohr (approved):

PR #3093, opened 2023-09-25
use v4 store inside backgroundMgr service.

mki-c2c:
Instead of removing all legacy Bg managers (all occurences of ngeoBackgroundLayerMgr ), we map the v4 BG layer store inside the ngeo component.
This fixes all occurences of ngeoBackgroundLayerMgr.get
But the use of ngeoBackgroundLayerMgr.set has to be fixed case by case (for example in SearchController and Mymaps)


tkohr (approved):
If I saw correctly, this is already part of the print PR

PR #3094, opened 2023-09-25
Gslux 629 print

mki-c2c:
fix the print service by catching the custom URLs for MapLibreLayers
The correct bg layer is already obtained thanks to https://github.com/Geoportail-Luxembourg/geoportailv3/pull/3093


tkohr (approved):
Thanks @mki-c2c ! Does this cover vector, raster and thematic layers? Couldn't test it yet.

PR #3095, opened 2023-09-25
Gslux 651 search service

mki-c2c:
fix search service for background layers and other layers.

3d layers not working in search (but this is as in prod)

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/search/SearchController.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/search/SearchController.js
@@ -33,6 +33,9 @@ import Fuse from 'fuse.js';
 import { matchCoordinate } from '../CoordinateMatch'
 import olcsCore from 'olcs/core.js';
 
+import { useThemeStore, useBackgroundLayer, storeToRefs, watch } from "luxembourg-geoportail/bundle/lux.dist.mjs";
+
+
 /**
  * @ngInject
  * @constructor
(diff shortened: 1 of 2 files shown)


geoportal/geoportailv3_geoportal/static-ngeo/js/search/SearchController.js
```diff
@@ -350,13 +355,13 @@ const exports = function($scope, $window, $compile,
     display: function(feature) {
       if (feature) {
         feature['dataset'] = 'backgroundLayers'
-        return this.gettextCatalog.getString(feature.get('label'))
+        return this.gettextCatalog.getString(feature.name)
```
tkohr, line 358: Why did this change?

tkohr (approved):
Thanks @mki-c2c ! Didn't see where you handle the thematic layers and couldn't properly test yet, but we might do this again in the 3D root branch.

geoportal/geoportailv3_geoportal/static-ngeo/js/search/SearchController.js
mki-c2c: the v4 store provides "Layer" objects which have the name attribute, whereas the legacy code used full featured OL layers, where the "label" property was set at creation.
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportal/geoportailv3_geoportal/static-ngeo/js/GetWmtsLayerFactory.js#L105
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportal/geoportailv3_geoportal/static-ngeo/js/GetWmsLayerFactory.js#L57
https://github.com/Geoportail-Luxembourg/geoportailv3/blob/master/geoportal/geoportailv3_geoportal/static-ngeo/js/Themes.js#L159

PR #3097, opened 2023-10-12
GSLUX-658: Fix display bottom right btn controls for FF

AlitaBernachot:
### JIRA issue

https://jira.camptocamp.com/browse/GSLUX-658

### Description

Right bottom button controls (offline mode, 3D, and info) were not visible in Firefox.
Now there are visible on Chrome, FF and Edge.


tkohr (approved):
Good catch @AlitaBernachot ! Displays well in FF (118.0.2) now.

PR #3137, opened 2024-04-29
GSLUX-663: Reload themes on user login and logout

AlitaBernachot:
https://jira.camptocamp.com/browse/GSLUX-663

Force reload and refresh themes in themes selector each time the user logs in or logs out.


tkohr (approved):
Thanks, @AlitaBernachot ! I didn't test it, but logic makes sense to me.

PR #3138, opened 2024-05-03
GSLUX-656: Lost background when using bg switcher

AlitaBernachot:
It looks like it is fixed when lux lib is up to date (use last commit `1e577fd5faf3d0426a44ce0733c95281ceebd199`).


AlitaBernachot:
Closing PR, bg issue has been fixed, we just need to use the correct hash for Lux lib.

PR #3139, opened 2024-05-06
GSLUX-601: Fix rebase

tkohr:
to call setThemes only once


mki-c2c (approved):
work fine on my machine

PR #3149, opened 2024-06-04
GSLUX-698: FIX init bg config AFTER urls init

AlitaBernachot:
### JIRA issue

https://jira.camptocamp.com/browse/GSLUX-698

### Description

Add init bg configs function, this initialization must be made AFTER vectortiles url initialization (this a a patch for v3)


tkohr (approved):
Thanks, @AlitaBernachot !

PR #3150, opened 2024-06-06
GSLUX-667: Offline mode

AlitaBernachot:
https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/pull/107


tkohr (commented):
Waooh this is some big work, especially in the v4 PR, thanks @AlitaBernachot! 

I did not dive into all details of the code, but I've noticed one strange behavior when testing: When having created a mymaps with a geometry (using first orthophoto layer as background), saving the data for offline and then switching to offline mode, the background layer switches to the second roads layer and the geometry disappears. (The background switching seems to be due to the mymaps not having a background defined, but I'm unable to save one to it.) If I interrupt the connection before switching to offline mode the background layer does not switch, but the geometry still disappears. 

Besides this and the potentially missing vector layers caching, it works really fine. Bravo!

tkohr (approved):
Thanks for the fix @AlitaBernachot! It works really well now! Even a geometry created offline is kept when switching back to online, which is not the case on prod I think.

PR #3151, opened 2024-06-11
Gslux 698 fix mvt urls

mki-c2c:
fix problem with style URL registration

### DIFF
--- a/env.default
+++ b/env.default
@@ -50,4 +50,6 @@ GDAL_DISABLE_READDIR_ON_OPEN=TRUE
 no_proxy=webpack_dev_server
 
 PROXYWMSURL=https://map.geoportail.lu/ogcproxywms
+UPLOAD_VT_STYLE_URL=https://migration.geoportail.lu/uploadvtstyle
+DELETE_VT_STYLE_URL=https://migration.geoportail.lu/deletevtstyle
 VECTORTILESURL=https://vectortiles-staging.geoportail.lu
--- a/geoportal/geoportailv3_geoportal/__init__.py
+++ b/geoportal/geoportailv3_geoportal/__init__.py
@@ -476,6 +476,9 @@ def main(global_config, **settings):
     config.add_route('upload_vt_style', '/uploadvtstyle')
     config.add_route('delete_vt_style', '/deletevtstyle')
     config.add_route('get_vt_style', '/getvtstyle')
+    config.add_route('upload_permalink_style', '/uploadpermalinkstyle')
+    config.add_route('delete_permalink_style', '/deletepermalinkstyle')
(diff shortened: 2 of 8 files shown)


env.default
```diff
@@ -50,4 +50,6 @@ GDAL_DISABLE_READDIR_ON_OPEN=TRUE
 no_proxy=webpack_dev_server
 
 PROXYWMSURL=https://map.geoportail.lu/ogcproxywms
+UPLOAD_VT_STYLE_URL=https://migration.geoportail.lu/uploadvtstyle
+DELETE_VT_STYLE_URL=https://migration.geoportail.lu/deletevtstyle
```
tkohr, line 54: Why don't we need `GET_VT_STYLE_URL` here?

geoportal/geoportailv3_geoportal/__init__.py
```diff
@@ -476,6 +476,9 @@ def main(global_config, **settings):
     config.add_route('upload_vt_style', '/uploadvtstyle')
     config.add_route('delete_vt_style', '/deletevtstyle')
     config.add_route('get_vt_style', '/getvtstyle')
+    config.add_route('upload_permalink_style', '/uploadpermalinkstyle')
+    config.add_route('delete_permalink_style', '/deletepermalinkstyle')
+    config.add_route('get_permalink_style', '/getpermalinkstyle')
```
tkohr, line 481: Why do we need to introduce the  `...permalink...` entry points for the v4 additionally to the other ones?

geoportal/geoportailv3_geoportal/static-ngeo/js/apps/MainController.js
```diff
@@ -465,7 +466,7 @@ const MainController = function(
     ngeoDownload, appMvtStylingService, ngeoDebounce, geonetworkBaseUrl, appBlankLayer,
     proxyWmsUrl,
     httpsProxyUrl,
-    getvtstyleUrl, uploadvtstyleUrl, deletevtstyleUrl, vectortilesUrl) {
+    getvtpermalinkUrl, uploadvtpermalinkUrl, deletevtpermalinkUrl, vectortilesUrl) {
```
tkohr, line 469: Same question as above: Why do we need to use the  `...permalink...` entry points here instead of the existing ones?

tkohr (approved):
Thanks for the fix @mki-c2c ! Just left some comments inline for my understanding.

geoportal/geoportailv3_geoportal/__init__.py
mki-c2c: - permalink storage (for permalink sharing and maputnik style editor) no sync with tile server raster renderer
- tile server ( only upload and delete) raster tiles are rendered for 3D and printing

PR #3166, opened 2024-07-08
GSLUX-713: put necessary offline files into cache also in prod

mki-c2c:
bugfix offline files


AlitaBernachot (approved):
Did not test but looks good, thx!

PR #3167, opened 2024-07-08
GSLUX-701: Missing text "No layer selected" when offline

AlitaBernachot:
### JIRA issue

https://jira.camptocamp.com/browse/GSLUX-701

### Description

- Add missing text "No layer selected" when user switches to offline mode
- in v4: update icons and add new scholl theme (after merging main onto integration-vue)

NB. This PR needs v4 fix : https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/pull/117

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/Mymaps.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/Mymaps.js
@@ -25,7 +25,7 @@ import olStyleStyle from 'ol/style/Style.js';
 import {get as getProj, transform} from 'ol/proj.js';
 import olcsCore from 'olcs/core.js';
 
-import { useMapStore, useThemeStore, useBackgroundLayer, useLayers, useThemes, storeToRefs, watch } from "luxembourg-geoportail/bundle/lux.dist.js";
+import { useAppStore, useMapStore, useThemeStore, useBackgroundLayer, useThemes, storeToRefs, watch } from "luxembourg-geoportail/bundle/lux.dist.js";
 
 /**
  * @constructor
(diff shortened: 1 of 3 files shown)


AlitaBernachot:
> (which btw should even be displayed without any MyMaps open, when testing on prod, right?)

Yes, right! I didn't see that, will fix this.

> When testing, I'm not able to see the message when switching to offline mode

weird... Did you test http://localhost:8080/dev/main.html or just http://localhost:8080/? because I did not commit the bundle on purpose. But I will check again.

tkohr:
Ok so I didn't look in the right place of the UI with MyMaps it works :-)

geoportal/geoportailv3_geoportal/static-ngeo/js/Mymaps.js
```diff
@@ -485,6 +485,9 @@ exports.prototype.setMapId = function(mapId) {
   this.stateManager_.updateState({
     'map_id': this.mapId_
   });
+
+  // v4 Set map id MyMaps in lib
+  useAppStore().setMapId(this.mapId_);
```
tkohr, line 490: If I get it right this is not used anymore for the message of this PR, but kept for later, right?

tkohr (approved):
Did not retest it since I need to rebuild, but changes look good to me, thanks!

geoportal/geoportailv3_geoportal/static-ngeo/js/Mymaps.js
AlitaBernachot: yes indeed

PR #3174, opened 2024-07-23
GSLUX-720: Fix reading permalink containing 3d_enabled

tkohr:
This PR should allow opening permalinks in 3D mode correctly (eg. https://migration.geoportail.lu/theme/main?version=3&X=662398&Y=6360606&zoom=18&lang=fr&rotation=0&bgLayer=streets_jpeg&crosshair=false&layers=&opacities=&time=&streets_jpeg=&serial=&3d_layers=ACT2022_BD_L_BATI3D_LOD2%2CVDL2022_LOD2_IMPORT_old%2CBridges_LOD2%2CACT2019_LiDAR_Bridges_LOD1%2CACT2019_LiDAR_Vegetation&3d_enabled=true&3d_lon=5.95042&3d_lat=49.49531&3d_elevation=696&3d_heading=360.000&3d_pitch=-39.984).

For the fix, the `appLayerPermalinkManager.initBgLayers_()` which returned an empty Promise is not called anymore. The included logic of the function already does not seem to be used anymore and was thus not migrated further into v4. The part regarding the `tourisme` theme could not be reproduced on production, and opacity logic only seems relevant for v2.

### DIFF
--- a/geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
+++ b/geoportal/geoportailv3_geoportal/static-ngeo/js/LayerPermalinkManager.js
@@ -594,55 +594,56 @@ exports.prototype.init = function(scope, map, selectedLayers) {
   this.initialized_ = true;
 };
 
-exports.prototype.initBgLayers_ = function() {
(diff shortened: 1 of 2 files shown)


AlitaBernachot:
LGTM, thanks for the fix!

PR #3175, opened 2024-07-25
GSLUX-715: Restrict zoom in offline mode

tkohr:
This PR restricts offline mode to the thematic layers zoom levels if a thematic layer is present. In this case it does not include zooms for background layers on save in order to take the first and last zoom levels from the thematic layers when restoring via getItem('offline_content').

This seems a little like a workaround since this condition is not necessary on prod. 

That being said, the missing zoom restrictions actually also seem to be present on prod when using a thematic layer with the orthophotos as a baselayer.


AlitaBernachot:
LGTM! thx

PR #3178, opened 2024-07-29
GSLUX-726: Highlight cadastral parcel on search

tkohr:
This PR should fix:
- highlighting a cadastral parcel selected in the search
- adding layers from corresponding theme (cherry-picked from master and not tested locally)
- not removing cadastral parcel layer on a new cadastral search if one is already active

PR can be tested locally with tweak plugin for example using a `http://localhost:8080/fulltextsearch?query=8&limit=8` response from prod that includes a parcel.


AlitaBernachot (approved):
Tested with tweak plugin, working great. Thx for the fix!

PR #3179, opened 2024-07-29
GSLUX-722: Fix embedded mode

AlitaBernachot:
### JIRA issue

https://jira.camptocamp.com/browse/GSLUX-722

### Description

- Add new functionality to v4: embedded mode (just add param `?embedded=true` in url)
- Force resize map when embedded mode, with a timer for v3 (because of angular component, wait until they have finish rendering header and footer that will disappear, otherwise, there is an unwanted gutter). For v4, no timer needed.
- 
NB. This PR needs https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/pull/125


tkohr (approved):
Thanks @AlitaBernachot !

PR #3180, opened 2024-07-30
Gslux 635 use v4 release bundle

mki-c2c:
needs the published bundles in https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/pull/128


geoportal/Dockerfile
```diff
@@ -45,17 +46,22 @@ RUN node --version
...
     ln -s /usr/lib/node_modules/@fortawesome/fontawesome-free/webfonts/fa-solid-900.svg /usr/lib/node_modules/@fortawesome/fontawesome-free/webfonts/fa-solid-900.svg#fontawesome
 
-
 # RUN make checks
 RUN make build
 
+RUN curl -L https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/releases/download/$(cat v4_release.txt)/bundle.tgz > /tmp/bundle.tgz && \
+    tar -xzvf /tmp/bundle.tgz -C /tmp && \
+    mkdir -p /usr/lib/node_modules/luxembourg-geoportail && \
+    mv /tmp/bundle/ /usr/lib/node_modules/luxembourg-geoportail
```
tkohr, line 63: I'm not a big fan of manipulating the npm dependencies "manually" here, as the `package.json` is already there to manage such things. Maybe we could find a way to continue making the bundle available that is directly accessible for the `package.json` (in a dedicated branch or via npm for example).

mki-c2c: I agree with you.
There are three arguments in favor of this approach, though:
- it's "less bad" than before with the bundle inside the repo and the whole source files installed and unused
- it's only for the time of the backport while v3 and v4 coexist. And it's by far not the only hack for this time
- in case of a V4 change, the docker image only has to be rebuilt for the last stages, the (long) installation of basic V3 npm dependencies can be skipped

so in the end I do not feel too bad with this "hack", but I'll check if we can build a "fake" npm package to install via the git hash...

AlitaBernachot: Agree with @tkohr

geoportal/Dockerfile
```diff
@@ -36,7 +36,8 @@ COPY geoportailv3_geoportal/static-ngeo /app/geoportailv3_geoportal/static-ngeo
 RUN rm -rf /usr/lib/node_modules/ngeo
 RUN mv /app/geoportailv3_geoportal/static-ngeo/ngeo /usr/lib/node_modules/ngeo
 
-COPY . /app
+RUN mkdir -p /app
+COPY ./package.json /app
 
 # jsapi generation
 ADD ./jsapi /etc/apiv4/
```
tkohr, line 43: I think we can get rid of the diff of the `Dockerfile` completely, as the `npm install` in between the two changes only concerns the JS API, if I see correctly. The `package.json` for the app is actually already copied in line 28.

mki-c2c: the main point of this diff is removing the `COPY . /app` and putting it after the jsapi build
but yes, we can remove 
```
RUN mkdir -p /app
COPY ./package.json /app
```

tkohr: Ok I don't get the advantage of moving `COPY . /app`, neither as the `WORKDIR` changes, but I might be missing something.

mki-c2c: By moving the cp . as late as possible we avoid all unnecessary docker build steps (npm install for jsapi and build jsapi) if the source code changes
This is just a build optimization which may reduce the image build time if something changes in app

We can leave the Dockerfile as is if that is more comfortable for you.

tkohr: Alright, I'm fine with moving the `COPY . /app`

tkohr (approved):
Thanks @mki-c2c ! Don't forget to remove copying the `package.json` twice and to update the tag, once the v4 PR is merged.

PR #3181, opened 2024-07-30
fix correct update of permalink

mki-c2c:
I believe this bug is related to the problem of bgLayer missing sometimes
This fix resolves a conflict in permalink update between v3 and v4 leading to js errors and incorrect styling.
Once trapped in a pad permalink / localstorage combination, bad things happen.

Unfortunately, it is again impossible to reproduce the reported bug, so we should merge the fix to check if evrything is going fine in the deployed app


tkohr (approved):
Thanks @mki-c2c !

PR #3193, opened 2024-10-15
GSLUX-758: Fix cond when adding a layer when bg is blank

AlitaBernachot:


tkohr (approved):
Thanks, @AlitaBernachot! Not tested, but change looks good regarding error message.

PR #3202, opened 2024-12-01
GSLUX-767: Use new v4 profile component in routing

AlitaBernachot:
https://jira.camptocamp.com/browse/GSLUX-767

Use new v4 Profile Component in Routing.

PR to be used: https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/pull/176

NB. Routing is not available  in dev, use chrome extension (eg. Tweak) to modify api responses. Also, units in dev does not correspond to production, you may again tweak de api response to get the real values (did not find a way to configure units in v3.


tkohr (approved):
Works with corresponding PR of v4 and mocked data. Thanks!

PR #3206, opened 2024-12-09
GSLUX-770: Fix draw and layers in permalink v3

AlitaBernachot:
### JIRA issue

https://jira.camptocamp.com/browse/GSLUX-770

### Description

Use storage helper from v4 to update correctly features in url without removing newly added layers.

- [x]  Update package.json before merging https://github.com/Geoportail-Luxembourg/luxembourg-geoportail/pull/179


tkohr (approved):
Good find, thanks for the fix @AlitaBernachot ! Adding layers and drawing works as expected in the URL now. I'm wondering if there are still other cases we haven't thought of, where `ngeoLocation` and the `state persistor` may be conflicting.

mki-c2c (approved):
works fine for some manual test cases

thanks for the fix

PR #3354, opened 2026-08-27
GSLUX-818: add is_visible flag in feature table

zurfjereluhmie:
Add a new flag to store whether a feature should be hidden or not in the UI


tkohr (approved):
Change LGTM and works on local dev db.

